PR #7, opened 2022-07-07
Update IDP sync environment variables

meyerjrr:
Add pass through for idp settings for sync handlers


cmd/lambda/approvals/handler.go
```diff
@@ -70,11 +72,19 @@ func buildHandler() (*Lambda, error) {
 		return nil, err
 	}
 
+	var sync deploy.Identity
+	err = json.Unmarshal([]byte(cfg.IdentitySettings), &sync)
+	if err != nil {
+		panic(err)
```
JoshuaWilkes:
```suggestion
		return nil, err
```

cmd/server/main.go
```diff
@@ -95,11 +95,19 @@ func run() error {
 		return err
 	}
 
+	var sync deploy.Identity
+	err = json.Unmarshal([]byte(cfg.IdentitySettings), &sync)
+	if err != nil {
+		panic(err)
```
JoshuaWilkes:
```suggestion
		return err
```

JoshuaWilkes (changes requested):

JoshuaWilkes (approved):

PR #8, opened 2022-07-07
Azure idp

meyerjrr:


JoshuaWilkes (commented):
Looks good at the moment @meyerjrr

jordiup (approved):

chrnorm (changes requested):
Looks like the Azure SDK forces a minimum of Go 1.18 (https://github.com/Azure/azure-sdk-for-go/issues/17472) - currently this project is on 1.17.

@meyerjrr rather than using the Azure SDK, could you try and call the Microsoft Graph APIs using the stdlib `net/http`? As we add more providers, the more that we can use the stdlib for is better anyway because it will limit the amount of third party dependencies we're pulling in to Granted.

[Here's a tutorial](https://blog.logrocket.com/making-http-requests-in-go/) on using `net/http` for API requests - @JoshuaWilkes can pair with you if you need a hand making auth work with these requests.

You can use the [Microsoft Graph docs](https://docs.microsoft.com/en-us/graph/api/group-list?view=graph-rest-1.0&tabs=http#example-1-get-a-list-of-groups) which have lots of examples of how the HTTP request should look.

Alternatively we should be able to upgrade to 1.18 without any issues but I'd be very keen to try and drop the dependency as a first step.

PR #10, opened 2022-07-11
CF-1147: CheckComplete fn for gdeploy

jordiup:


cmd/gdeploy/commands/status.go
```diff
@@ -31,7 +31,17 @@ var StatusCommand = cli.Command{
 		}
 		o.PrintTable()
 
-		clio.Success("Your Granted deployment is online")
+		cc, err := dc.CheckComplete(ctx)
```
JoshuaWilkes: I'm thinking that this would be better if it just displayed the cloudformation status directly rather than checking the status and returning some strink.

DE = I can run the status command any time and see exactly what deployment status my cloudformation is in.

The gdeploy tool should be a convenience around using cloudformation directly but we shouldn't completely abstract it in my opinion .

I though you shoudl change the checkStatus method to be a getStackStatus which returns the cloudformation status

JoshuaWilkes (changes requested):

cmd/gdeploy/commands/status.go
jordiup: Hey @JoshuaWilkes, appreciate the review, I've pushed another commit which I think should address this :)

cmd/gdeploy/commands/status.go
```diff
@@ -31,7 +31,17 @@ var StatusCommand = cli.Command{
 		}
 		o.PrintTable()
 
-		clio.Success("Your Granted deployment is online")
+		ss, err := dc.GetStackStatus(ctx)
+
+		if err != nil {
+			return err
+		}
+		if ss == "CREATE_COMPLETE" {
```
JoshuaWilkes: @jordiup because cloudformation can be in create_complete or update_complete

I propose that we don't have an if statement here and we simply just log a generic message with the status

```
clio.Info("Cloudformation stack status: %s", ss)
```

JoshuaWilkes (changes requested):

JoshuaWilkes (approved):

PR #14, opened 2022-07-11
add json defaults for identity settings

JoshuaWilkes:
Fixes an issue where the dashboard would endlessly redirect users due to the user directory sync setting JSON payload not being unmarshalled.


chrnorm (approved):
Looks good @JoshuaWilkes! To make this more defensive I've added an unmarshal function which handles the case where the variable is an empty string (in case we change config variables in future and forget to add the `default={}` tag)

PR #15, opened 2022-07-11
Add process for release candidate releasing

chrnorm:
new bucket `granted-test-releases-us-east-1` is public which simplifies RC deployment testing

separate 'dev' and 'rc' folders:
- dev folder is for each commit
- rc folder is for RCs only

This also swaps the release bucket region from `ap-southeast-2` to `us-east-1` - the only motivation to use `us-east-1` is that we have our testing Cognito user pool there and that it's the same region as GitHub Actions runs in which should minimise latency. In future we can add other region buckets like `granted-test-releases-ap-southeast-2`.

Additionally, this PR adds support for using custom URLs in the `release` field in our deployment YAML config. This will allow us to point a deployment at a release candidate.

```yaml
deployment:
  stackName: ExampleStack
  account: "123456789012"
  region: us-east-1
  release: https://granted-test-releases-us-east-1.s3.amazonaws.com/rc/v0.1.1-rc1
```

We can use this to ensure that it's possible to upgrade from an existing release to the next RC, as some operations in CloudFormation can cause errors during stack updates (such as adding multiple GSIs at once)

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -118,8 +118,8 @@ jobs:
       - name: Configure AWS Credentials
         uses: aws-actions/configure-aws-credentials@v1
         with:
-          aws-region: us-east-1
+          aws-region: us-west-2
           role-to-assume: "${{ secrets.AWS_TESTING_ROLE_ARN }}"
 
       - name: Release CloudFormation
-        run: go run mage.go -v release:production granted-test-releases-us-east-1 dev/${{ github.sha }}
+        run: go run mage.go -v release:production granted-test-releases-west-2 dev/${{ github.sha }}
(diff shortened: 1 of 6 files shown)


.github/workflows/deploy.yml
```diff
@@ -118,8 +118,8 @@ jobs:
...
         uses: aws-actions/configure-aws-credentials@v1
         with:
-          aws-region: ap-southeast-2
+          aws-region: us-west-2
           role-to-assume: "${{ secrets.AWS_TESTING_ROLE_ARN }}"
 
-      - name: Release Production Assets
-        run: go run mage.go -v release:production granted-internal-testing-releases ${{ github.sha }}
+      - name: Release CloudFormation
+        run: go run mage.go -v release:production granted-test-releases-west-2 dev/${{ github.sha }}
```
JoshuaWilkes: is this a typo ?
```suggestion
        run: go run mage.go -v release:production granted-test-releases-us-west-2 dev/${{ github.sha }}
```

JoshuaWilkes (approved):

PR #19, opened 2022-07-12
Add json defaults for slack and access handler

JoshuaWilkes:
This PR ensures that we have the correct default values for slack and access handler config.
I have added an unmarshaller for slackconfig and a default check for the access handler provider config

### DIFF
--- a/accesshandler/pkg/config/providers.go
+++ b/accesshandler/pkg/config/providers.go
@@ -33,7 +33,7 @@ func ReadProviderConfig(ctx context.Context, runtime string) ([]byte, error) {
 	if !ok {
 		return nil, errors.New("PROVIDER_CONFIG environment variable not set")
 	}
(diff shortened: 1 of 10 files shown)


accesshandler/pkg/config/providers.go
```diff
@@ -33,6 +33,10 @@ func ReadProviderConfig(ctx context.Context, runtime string) ([]byte, error) {
 	if !ok {
 		return nil, errors.New("PROVIDER_CONFIG environment variable not set")
 	}
+	// ensure that if the env var is set but is an epty string, that we replace it defensively with an empty json object to prevent 500 errors
```
jordiup:
```suggestion
	// ensure that if the env var is set but is an empty string, we replace it defensively with an empty json object to prevent 500 errors
```

jordiup (approved):
Looks good Josh

PR #20, opened 2022-07-12
CF-1146: Check credentials are valid

jordiup:
Running `gdeploy` commands now checks `MustGetCurrentAccountID` to ensure session is valid prior to running dependent commands 

image

### DIFF
--- a/cmd/gdeploy/commands/create.go
+++ b/cmd/gdeploy/commands/create.go
@@ -19,7 +19,7 @@ var CreateCommand = cli.Command{
 		ctx := c.Context
 
 		// Ensure aws account session is valid
-		deploy.MustGetCurrentAccountID(ctx, deploy.WithWarnExpiryIfWithinDuration(time.Minute))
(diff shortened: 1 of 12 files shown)


JoshuaWilkes:
Nice one Jordi, does this add any noticeable latency to the commands?

If it does, we could look at adding another function MustHaveCredentials() which would be the first part of MustGetCurrentAccountID except that it doesn't make the call to aws sts get-caller-identity

cmd/gdeploy/commands/create.go
```diff
@@ -16,6 +18,9 @@ var CreateCommand = cli.Command{
 	Action: func(c *cli.Context) error {
 		ctx := c.Context
 
+		// Ensure aws account session is valid
+		deploy.MustGetCurrentAccountID(ctx, deploy.WithWarnExpiryIfWithinDuration(time.Minute))
```
JoshuaWilkes:
```suggestion
		deploy.MustGetCurrentAccountID(ctx, deploy.WithWarnExpiryIfWithinDuration(time.Minute * 10))
```

JoshuaWilkes (approved):

PR #22, opened 2022-07-12
Use correct region when working in gdeploy commands

JoshuaWilkes:
This PR seeks to standardise and improve the way we use AWS config and deployment config across gdeploy and the approvals backend / mage development commands

Commands can be wrapped with before functions which will configure requires aws config and or deployment config 
```
Commands: []*cli.Command{
	WithBeforeFuncs(&users.UsersCommand, RequireDeploymentConfig(), RequireAWSCredentials()),
}

```
deployment config will be available for commands via the following
```
dc, err := deploy.ConfigFromContext(ctx)
if err != nil {
	return err
}
```

aws config will be available via the following. The intention of the aws config function below is to use it everywhere in the codebase. it will either return the config which has been set in the context or a new default config.

Using this everywhere will make it easy to remember. 
```
cfg, err := cfaws.ConfigFromContextOrDefault(ctx)
if err != nil {
	return err
}
```

This Pr also adds clio.APIError
Which allows you to return an error which can print any combination of Log Error Warn Info without needing to call os.Exit 

``` go
return &clio.CLIError{
		Err: "Could not find AWS credentials.",
		Messages: []clio.Printer{
			&clio.InfoMsg{
				Msg: "Please export valid AWS credentials to run this command.",
			},
		},
	}


[e] Could not find AWS credentials. 
[i] Please export valid AWS credentials to run this command."
```

### DIFF
--- a/cmd/gdeploy/commands/sync/command.go
+++ b/cmd/gdeploy/commands/sync/command.go
@@ -18,7 +18,7 @@ var SyncCommand = cli.Command{
 		ctx := c.Context
 
 		// Ensure aws account session is valid
-		deploy.MustGetCurrentAccountID(ctx, deploy.WithWarnExpiryIfWithinDuration(time.Minute))
+		deploy.MustHaveAWSCredentials(ctx, deploy.WithWarnExpiryIfWithinDuration(time.Minute))
 
 		var cfg config.SyncConfig
 		_ = godotenv.Load()
--- a/cmd/gdeploy/commands/sync/command.go
+++ b/cmd/gdeploy/commands/sync/command.go
@@ -1,6 +1,14 @@
 package sync
 
 import (
+	"os"
+	"time"
+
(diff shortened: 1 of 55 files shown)


cmd/gdeploy/commands/sync/command.go
```diff
@@ -17,39 +10,15 @@ var SyncCommand = cli.Command{
...
-			return err
-		}
-
-		clio.Info("Starting sync")
-		err = syncer.Sync(ctx)
+		_, err := deploy.ConfigFromContext(ctx)
 		if err != nil {
 			return err
 		}
+		// _, err := dc.LoadOutput(ctx)
```
JoshuaWilkes: This command is awaiting reimplementation.
The previous version would have only work for developers with environment variables.

Expecting this will be implemented tomorrow

meyerjrr (approved):

PR #23, opened 2022-07-12
Jordi/cf 14 set up playwright

meyerjrr:
Playwright set up with auth injection
next steps 
- will be writing the actual tests
- CI 
Because of how Playwright sets up the user auth. Will need to work out how to write tests for users and admins separately. Can spec this up in a notion tomorrow.


web/package.json
```diff
@@ -53,12 +54,16 @@
     "@types/react-dom": "18.0.5",
     "@types/react-table": "^7.7.11",
     "@vitejs/plugin-react": "^1.3.2",
+    "aws-amplify": "^4.3.27",
```
chrnorm: can you remove this import? We have `@aws-amplify/auth` which you can use instead!

web/package.json
```diff
@@ -53,12 +54,16 @@
...
+    "dotenv": "^16.0.1",
     "eslint": "8.13.0",
     "eslint-config-next": "12.1.5",
+    "fs": "0.0.1-security",
```
chrnorm: `fs` is built in to Node so you shouldn't need to install it - we can remove this

chrnorm (dismissed):
quick couple of dependency suggestions from me

web/package.json
jordiup: Cheers Chris, amended

jordiup:
Going to merge this one in as our Desired Outcomes are met for this initial ticket :)

jordiup (approved):

PR #24, opened 2022-07-12
Josh/cf 11 gdeploy sync command should call the

JoshuaWilkes:
This is dependant on #22

### DIFF
--- a/pkg/clio/errors.go
+++ b/pkg/clio/errors.go
@@ -1,0 +1,76 @@
+package clio
+
+type Printer interface {
+	Print()
+}
+
+type PrintCLIErrorer interface {
+	PrintCLIError()
+}
+type CLIError struct {
+	// Err is a string to avoid puntuation linting
+	Err string
+	// setting ExcludeDefaultError to true will cause the PrintCLIError method to only print the contents of Messages
+	// by default PrintCLIError first prints the contents of Err, followed by the contents of Messages
+	ExcludeDefaultError bool
+	// Messages are items which implement the clio.Printer interface.
+	// when PrintCLIError is called, each of the messages Print() method is called in order of appearence in the slice
+	Messages []Printer
+}
+
+type LogMsg struct {
+	Msg string
+}
+
+func (m *LogMsg) Print() {
+	Log(m.Msg)
+}
+
+type InfoMsg struct {
+	Msg string
+}
+
+func (m *InfoMsg) Print() {
+	Info(m.Msg)
+}
+
+type WarnMsg struct {
+	Msg string
+}
+
+func (m *WarnMsg) Print() {
+	Warn(m.Msg)
+}
+
+type DebugMsg struct {
+	Msg string
+}
+
+func (m *DebugMsg) Print() {
+	Debug(m.Msg)
+}
+
+// Error implements the error interface. It uses the default message of the
+// wrapped error.
+func (e *CLIError) Error() string {
+	return e.Err
+}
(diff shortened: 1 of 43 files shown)


pkg/clio/errors.go
```diff
@@ -0,0 +1,76 @@
...
+// Error implements the error interface. It uses the default message of the
+// wrapped error.
+func (e *CLIError) Error() string {
+	return e.Err
+}
+
+// PrintCLIError prints the error message and then any messages in order from the slice
+// The indended use is to surface errors with useful messages then os.Exit without having to place os.Exit withing methods other than the cli main function
+//
+// err := CLIError{Err: errors.New("new error"), Messages: []Printer{&LogMsg{Msg:"hello world"}}}
```
meyerjrr, line 62: This might be a nitpick, but do we need to have the nested Msg struct on LogMsg. Would become a bit more readable if it was just a string. 
Unless you had plans to expand the LogMsg struct to include other fields.

meyerjrr (approved):

pkg/clio/errors.go
JoshuaWilkes: > This might be a nitpick, but do we need to have the nested Msg struct on LogMsg. Would become a bit more readable if it was just a string. Unless you had plans to expand the LogMsg struct to include other fields.



> This might be a nitpick, but do we need to have the nested Msg struct on LogMsg. Would become a bit more readable if it was just a string. Unless you had plans to expand the LogMsg struct to include other fields.

Good call Jack,



I had a go and looks like this will work just fine


``` go
&clio.CLIError{
Err: "Failed to load AWS credentials.",
Messages: []clio.Printer{
	clio.DebugMsg("Encountered error while loading default aws config: %s", err),
},

```

``` go
const (
	LogMsgType     MsgType = "LOG"
	InfoMsgType    MsgType = "INFO"
	SuccessMsgType MsgType = "SUCCESS"
	ErrorMsgType   MsgType = "ERROR"
	WarnMsgType    MsgType = "WARN"
	DebugMsgType   MsgType = "DEBUG"
)

type Msg struct {
	Msg  string
	Type MsgType
}

func (m Msg) Print() {
	switch m.Type {
	case InfoMsgType:
		Info(m.Msg)
	case SuccessMsgType:
		Success(m.Msg)
	case ErrorMsgType:
		Error(m.Msg)
	case WarnMsgType:
		Warn(m.Msg)
	case DebugMsgType:
		Debug(m.Msg)

	default:
		Log(m.Msg)
	}
}
func LogMsg(format string, a ...interface{}) Msg {
	return Msg{Msg: fmt.Sprintf(format, a...), Type: LogMsgType}
}
func InfoMsg(format string, a ...interface{}) Msg {
	return Msg{Msg: fmt.Sprintf(format, a...), Type: InfoMsgType}
}
func SuccessMsg(format string, a ...interface{}) Msg {
	return Msg{Msg: fmt.Sprintf(format, a...), Type: SuccessMsgType}
}
func ErrorMsg(format string, a ...interface{}) Msg {
	return Msg{Msg: fmt.Sprintf(format, a...), Type: ErrorMsgType}
}
func WarnMsg(format string, a ...interface{}) Msg {
	return Msg{Msg: fmt.Sprintf(format, a...), Type: WarnMsgType}
}
func DebugMsg(format string, a ...interface{}) Msg {
	return Msg{Msg: fmt.Sprintf(format, a...), Type: DebugMsgType}
}
```

PR #28, opened 2022-07-13
Jordi/cf 31 add GitHub action for playwright tests

jordiup:


.github/workflows/test.yml
```diff
@@ -95,3 +95,48 @@ jobs:
...
+        run: pnpm i
+        working-directory: web
+      - name: Install Playwright
+        run: pnpm run e2e-install
+        working-directory: web
+      - name: Run Playwright tests
+        run: pnpm run e2e
+        working-directory: web
+        env:
+          PLAYWRIGHT_TEST_BASE_URL: djmoxvh6xpra8.cloudfront.net/ # @TODO find a way to pass this between deployment
```
chrnorm:
```suggestion
```

we can remove this as we're only using `TESTING_DOMAIN`

chrnorm (changes requested):

.github/workflows/test.yml
```diff
@@ -95,3 +95,48 @@ jobs:
...
+          TESTING_DOMAIN: '${{ secrets.TESTING_DOMAIN}}' # @TODO find a way to pass this between deployment
```
chrnorm: `secrets.TESTING_DOMAIN` is okay for now - to get the benefit of testing in GitHub Actions we'll need to update this to point to the branch deployment URL (e.g. https://jordi-cf-31-add-github-action-fo.test.granted.run/). The SSL certs are currently broken here though so we need to fix those first!

.github/workflows/test.yml
```diff
@@ -95,3 +95,47 @@ jobs:
...
+        run: pnpm i
+        working-directory: web
+      - name: Install Playwright
+        run: pnpm run e2e-install
+        working-directory: web
+      - name: Run Playwright tests
+        run: pnpm run e2e
+        working-directory: web
+        env:
+          TESTING_DOMAIN: '${{ secrets.TESTING_DOMAIN}}' # @TODO find a way to pass this between deployment
```
chrnorm: I've now merged the changes from #32 which fixes the custom URLs up! You can now use:

```suggestion
          TESTING_DOMAIN: '${{ steps.branch.outputs.name }}.test.granted.run'
```

.github/workflows/test.yml
```diff
@@ -95,3 +95,47 @@ jobs:
...
+      # - name: Install Playwright
+      #   run: pnpm run e2e-install
+      #   working-directory: web
+      # - name: Run Playwright tests
+      #   run: pnpm run e2e
+      #   working-directory: web
+      #   env:
+      #     # This might depend on your test-runner/language binding
+      #     PLAYWRIGHT_TEST_BASE_URL: djmoxvh6xpra8.cloudfront.net/
+  playwright-tests:
```
chrnorm: Now that #32 is implemented, the playwright tests depend on the preview environment being created. Can you shift this job into `preview.yml` and make it dependent on the `deploy` job completing?

.github/workflows/test.yml
```diff
@@ -95,3 +95,47 @@ jobs:
...
+    # needs: [test-frontend]
+    timeout-minutes: 10
+    runs-on: ubuntu-latest
+    # if: github.event.deployment_status.state == 'success'
+    steps:
+      - uses: actions/checkout@v2
+      - name: Setup pnpm
```
chrnorm: we also need to get the branch name so we can use it below as ${{ steps.branch.outputs.name }}.test.granted.run'. You can do so by adding the `common-fate/branch-name` action to this workflow.

```
      - uses: common-fate/branch-name@v1.1.2
        name: Get branch name
        id: branch
        with:
          max-length: 32
```

chrnorm (changes requested):
Now that #32 is implemented we can get this up and running.

PR #31, opened 2022-07-13
skip user prompt if CI env var is true

chrnorm:


cmd/gdeploy/commands/update.go
```diff
@@ -23,6 +25,12 @@ var UpdateCommand = cli.Command{
 		clio.Info("Deploying Granted Approvals %s", dc.Deployment.Release)
 		clio.Info("Using template: %s", dc.CfnTemplateURL())
 		confirm := c.Bool("confirm")
+
+		if os.Getenv("CI") == "true" {
```
JoshuaWilkes, line 29: I wonder, can this be set as an environment flag directly in the cli flags?

JoshuaWilkes (approved):

PR #35, opened 2022-07-14
Josh/cf 27 create data model for request event

JoshuaWilkes:
image

Work in progress, testing this approach for audit logs on requests.
RCF


pkg/eventhandler/eventhandler.go
```diff
@@ -50,28 +51,36 @@ func (n *EventHandler) HandleGrantEvent(ctx context.Context, log *zap.SugaredLog
...
+	if oldStatus == newStatus {
+		// I anticipate that this would be succeptible to a race condition, recoverable if the eventbridge retries the event handler
+		// this is because the grant events are sourced from the access handler prior to the request being saved to dynamodb on creation
+		// we could solve this by saving the request to the DB prior to making the call to the access handler?
+		if event.DetailType == gevent.GrantCreatedType {
+			requestEvent := access.NewGrantCreatedEvent(gq.Result.ID, time.Now(), nil)
+			log.Infow("inserting request event for grant created")
+			return n.db.Put(ctx, &requestEvent)
+		}
+		return nil
```
chrnorm: I found this a bit confusing to read - we could make it simpler by reducing the levels of nesting

ie

```go
if oldStatus != newStatus {
  // emit status change event
}

if event.DetailType == gevent.GrantCreatedType {
  // emit grant created event
}
```

chrnorm:
Agreed on the approach and the API design here 👍

JoshuaWilkes:
With all events 
image

chrnorm:
Looks good - a UI nitpick on this, can you use lowercase instead of all caps for the status? e.g. "pending" rather than "PENDING"

ellie-narducci (approved):

PR #39, opened 2022-07-15
Add e2e access rule tests

meyerjrr:


web/src/components/forms/access-rule/steps/FormStep.tsx
```diff
@@ -94,6 +94,7 @@ const BottomActionButtons: React.FC<{
     <Flex w="100%" justify={"left"}>
       {showNext && (
         <Button
+          id="form-step-next-button"
```
chrnorm, line 97: @meyerjrr can you please switch these attributes to be `data-testid` instead of `id`? My thinking is:
- The 'id' field is used for styling and navigation whereas what we're implementing is for testing only
- If nested components have the same 'id' the browser will usually throw errors in the console

chrnorm (changes requested):

PR #43, opened 2022-07-18
Cli fixes

meyerjrr:
Add check in AWS creds before, to check that the assumed role is for the same account as what is in the granted config
Fix bug causing account not to be saved to the config
Fixed UI in CLI when updating changeset has no changes
Check the status when creating stack and only display table + instructions after success

### DIFF
--- a/cmd/gdeploy/commands/create.go
+++ b/cmd/gdeploy/commands/create.go
@@ -56,7 +56,7 @@ var CreateCommand = cli.Command{
 Check out the next steps in our getting started guide for more information: https://docs.comonfate.io/granted-approvals/getting-started/deploying
 `)
 		} else {
-			clio.Warn("Your Granted deployment update ended in status %s", status)
+			clio.Warn("Creating your Granted deployment failed with a final status: %s", status)
 			return nil
 		}
 
--- a/cmd/gdeploy/commands/create.go
+++ b/cmd/gdeploy/commands/create.go
@@ -33,7 +33,7 @@ var CreateCommand = cli.Command{
 			confirm = true
 		}
 
-		_, err = dc.DeployCloudFormation(ctx, confirm)
+		status, err := dc.DeployCloudFormation(ctx, confirm)
 		if err != nil {
 			return err
 		}
@@ -42,10 +42,12 @@ var CreateCommand = cli.Command{
 		if err != nil {
 			return err
 		}
-		o.PrintTable()
 
-		clio.Success("Your Granted deployment has been created")
-		clio.Info(`Here are your next steps to get started:
+		if status == "CREATE_COMPLETE" {
+			clio.Success("Your Granted deployment has been created")
+			o.PrintTable()
+
+			clio.Info(`Here are your next steps to get started:
 
   1) create an admin user so you can log in: 'gdeploy users create --admin -u YOUR_EMAIL_ADDRESS'
   2) add an Access Provider: 'gdeploy provider add'
@@ -53,6 +55,10 @@ var CreateCommand = cli.Command{
 
 Check out the next steps in our getting started guide for more information: https://docs.comonfate.io/granted-approvals/getting-started/deploying
 `)
+		} else {
+			clio.Warn("Your Granted deployment update ended in status %s", status)
+			return nil
+		}
 
 		return nil
 	},
--- a/cmd/gdeploy/main.go
+++ b/cmd/gdeploy/main.go
@@ -176,8 +176,9 @@ func RequireAWSCredentials() cli.BeforeFunc {
 			return clio.NewCLIError("Failed to call AWS get caller identity. (maybe you need to assume a role first?)", clio.DebugMsg(err.Error()))
 		}
 
+		overwrite := c.Bool("overwrite")
 		//check to see that account number in config is the same account that is assumed
-		if identity.Account != &dc.Deployment.Account {
+		if identity.Account != &dc.Deployment.Account && !overwrite {
 			return clio.NewCLIError(fmt.Sprintf("Deployment account mismatched assumed account: %s verses %s", dc.Deployment.Account, aws.StringValue(identity.Account)))
 		}
 		c.Context = cfaws.SetConfigInContext(ctx, cfg)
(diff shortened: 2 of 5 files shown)


cmd/gdeploy/commands/create.go
```diff
@@ -42,17 +42,23 @@ var CreateCommand = cli.Command{
...
+			clio.Info(`Here are your next steps to get started:
 
   1) create an admin user so you can log in: 'gdeploy users create --admin -u YOUR_EMAIL_ADDRESS'
   2) add an Access Provider: 'gdeploy provider add'
   3) visit the web dashboard: 'gdeploy dashboard open'
 
 Check out the next steps in our getting started guide for more information: https://docs.comonfate.io/granted-approvals/getting-started/deploying
 `)
+		} else {
+			clio.Warn("Your Granted deployment update ended in status %s", status)
```
JoshuaWilkes: Word this specific to creating
```suggestion
			clio.Warn("Creating your Granted deployment failed with a final status %s", status)
```

cmd/gdeploy/main.go
```diff
@@ -165,14 +166,21 @@ func RequireAWSCredentials() cli.BeforeFunc {
...
-		_, err = stsClient.GetCallerIdentity(ctx, &sts.GetCallerIdentityInput{})
+		identity, err := stsClient.GetCallerIdentity(ctx, &sts.GetCallerIdentityInput{})
 		if err != nil {
 			var ae smithy.APIError
 			// the aws sdk doesn't seem to have a concrete type for ExpiredToken so instead we check the error code
 			if errors.As(err, &ae) && ae.ErrorCode() == "ExpiredToken" {
 				return clio.NewCLIError("AWS credentials are expired.", clio.LogMsg("Please export valid AWS credentials to run this command."))
 			}
-			return clio.NewCLIError("Failed to call AWS get caller identity", clio.DebugMsg(err.Error()))
+			return clio.NewCLIError("Failed to call AWS get caller identity. (maybe you need to assume a role first?)", clio.DebugMsg(err.Error()))
```
JoshuaWilkes:
```suggestion
			return clio.NewCLIError("Failed to call AWS get caller identity.", clio.LogMsg("Please export valid AWS credentials to run this command."), clio.DebugMsg(err.Error()))
```

cmd/gdeploy/main.go
```diff
@@ -165,14 +166,21 @@ func RequireAWSCredentials() cli.BeforeFunc {
...
+		}
+
+		//allow the overwrite if overwrite is set
+		overwrite := c.Bool("overwrite")
```
JoshuaWilkes: I don't think we need to have an override here at all, users should be deploying with the correct account every time.
You cannot do anything with the wrong account
```suggestion
```

cmd/gdeploy/main.go
```diff
@@ -165,14 +166,21 @@ func RequireAWSCredentials() cli.BeforeFunc {
...
+		//check to see that account number in config is the same account that is assumed
+		if *identity.Account != dc.Deployment.Account && !overwrite {
```
JoshuaWilkes:
```suggestion
		if *identity.Account != dc.Deployment.Account {
```

cmd/gdeploy/main.go
```diff
@@ -165,14 +166,21 @@ func RequireAWSCredentials() cli.BeforeFunc {
...
+			return clio.NewCLIError(fmt.Sprintf("Deployment account mismatched assumed account: %s verses %s", dc.Deployment.Account, aws.StringValue(identity.Account)))
```
JoshuaWilkes: I think we should avoid "assumed" terminology in this CLI, because our customers may be using any means to get their credentials. It will be more clear to just refer to them as AWS credentials 
```suggestion
			return clio.NewCLIError(fmt.Sprintf("AWS account in your deployment config %s does not match the account of your current AWS credentials %s", dc.Deployment.Account, *identity.Account), clio.LogMsg("Please export valid AWS credentials for account %s to run this command.", *identity.Account))
```

cmd/gdeploy/main.go
```diff
@@ -9,6 +9,7 @@ import (
 	"time"
 
 	"github.com/aws/aws-sdk-go-v2/service/sts"
+	"github.com/aws/aws-sdk-go/aws"
```
JoshuaWilkes: Be careful when importing aws that you import aws-sdk-go-v2 and not v1

```suggestion
```

JoshuaWilkes (changes requested):
I have requested some changes mostly to wording of the log messages

cmd/gdeploy/main.go
meyerjrr: This was to allow for the scenario where you want to overwrite your config with a different account. If you had creds for a different account it would never let you override.

The only way would be to delete the config and re-init to bypass this I think

JoshuaWilkes: In this case the customer would be creating a whole new deployment, they could change the account number manually in their config.

It better for us to help prevent this as much as possible.

cmd/gdeploy/main.go
```diff
@@ -165,14 +165,21 @@ func RequireAWSCredentials() cli.BeforeFunc {
...
 			}
-			return clio.NewCLIError("Failed to call AWS get caller identity", clio.DebugMsg(err.Error()))
+			return clio.NewCLIError("Failed to call AWS get caller identity. ", clio.LogMsg("Please export valid AWS credentials to run this command."), clio.DebugMsg(err.Error()))
+		}
+
+		//allow the overwrite if overwrite is set
+		overwrite := c.Bool("overwrite")
+		//check to see that account number in config is the same account that is assumed
+		if *identity.Account != dc.Deployment.Account && !overwrite {
+			return clio.NewCLIError(fmt.Sprintf("AWS account in your deployment config %s does not match the account of your current AWS credentials %s", dc.Deployment.Account, *identity.Account), clio.LogMsg("Please export valid AWS credentials for account %s to run this command."))
```
JoshuaWilkes:
```suggestion
			return clio.NewCLIError(fmt.Sprintf("AWS account in your deployment config %s does not match the account of your current AWS credentials %s", dc.Deployment.Account, *identity.Account), clio.LogMsg("Please export valid AWS credentials for account %s to run this command.",*identity.Account))
```

JoshuaWilkes (approved):

PR #47, opened 2022-07-19
added docs for playwright

meyerjrr:


jordiup (approved):

JoshuaWilkes (approved):

docs/testing/testing.md
```diff
@@ -0,0 +1,42 @@
...
+USER_POOL_ID=""
+COGNITO_CLIENT_ID=""
+TESTING_DOMAIN=""
+```
+
+From here we will need to do some work to setup these variables. By this point you should have setup a dev environment using mage. If not checkout out [Deploying](./deploying.md).
+- Grab `USER_POOL_ID` and `COGNITO_CLIENT_ID` from your `.env` from your dev deployment, or find the cognito user pool in the console to get these ID's
+- Create a test user and test admin user in your cognito user pool (add the admin user to your admin group in cognito) 
+*You might need to update their login passwords upon making the accounts, this will be emailed to your email if setup correctly*. 
+    - To make extra accounts you can use your own email address with a suffix to make infinite amount of accounts. Eg. `test+1@commonfate.io` where +1 is the suffix
```
ellie-narducci, line 27: @meyerjrr I don't think the `+` email addresses work in this instance (bug to resolve later). I tried it and had issues. Perhaps we remove this part for now.

meyerjrr: Hmm, which part was not working for you. @jordiup and myself were able to setup testing accounts using ours.

ellie-narducci: Oh really, that's strange. I couldn't get access when using it. Maybe I did something odd. Please leave it as is then 💯

PR #50, opened 2022-07-20
Jordi/cf 41 updating request timing to a decimal

jordiup:


JoshuaWilkes:
@jordiup Can you please split this hours minutes component out and share it between the two forms

JoshuaWilkes:
# Access request form

- [x] I found this UI confusing initially because it looked like it was broken/ I couldn't change the time.

Can we have the buttons always visible instead of when focused 
image



- [x] I set the minutes to 12, then set the hours to 11, then the hours field value broke.

The form also doesn't validate that the hours and minutes need to be whole number values and it allowed me to submit, but the submit failed. No error was shown which is a seperate issue to this ticket.
image


- [x] Submitting the form with 10 hours 11 minutes caused a 500 error on the server because the duration was 0 seconds in the payload
image


- [x] When the maxDuration was 0 (which is a bug described below) I couldn't change the minutes field but I could change the hours field , I expected that if teh maxDuration is 0 or otherwise that both hours and minutes correctly validated/limited
- [x] as well, the increment button was not disabled event though I wan't able to increment the minutes
image
image



# Create Access Rule form
- [x] Same issue with 0 value for max duration on create access rule form 
image


- [x] BUG The access rule create or update api should enforce a minimum "maxDurationSeconds" of 60 seconds, the create and update form in the FE should enforce a 1 minute minimum too. You could implement this in the open API spec and the validation should happen automatically.

JoshuaWilkes:
- [x] the increment buttons should be disabled when the max value is reached
image

On create access rule, the form allows me to skip setting a time. The default value should be 1 minute on the form.

image


- [x] Should we have an indication to users that there is a max of 24 hours?

web/src/components/HoursMinutes.tsx
```diff
@@ -0,0 +1,155 @@
...
+
+  let maxH = maxDurationSeconds ? maxDurationSeconds / 3600 : 24;
+
+  const onBlurFn = () => {
+    if (hours != undefined && mins != undefined) {
+      const duration = hours * 60 * 60 + mins * 60;
+      // DE = when an out of bounds value is adjusted to maxSeconds, we need to update the hours and mins to match
+      if (maxDurationSeconds && duration > maxDurationSeconds) {
+        setValue(maxDurationSeconds);
+        let h = Number((maxDurationSeconds / 60 / 60).toString().split(".")[0]);
```
JoshuaWilkes: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/trunc

```suggestion
        let h = Math.trunc(maxDurationSeconds / 60 / 60);
```

web/src/components/HoursMinutes.tsx
```diff
@@ -0,0 +1,155 @@
...
+
+  let maxH = maxDurationSeconds ? maxDurationSeconds / 3600 : 24;
+
+  const onBlurFn = () => {
+    if (hours != undefined && mins != undefined) {
+      const duration = hours * 60 * 60 + mins * 60;
+      // DE = when an out of bounds value is adjusted to maxSeconds, we need to update the hours and mins to match
+      if (maxDurationSeconds && duration > maxDurationSeconds) {
+        setValue(maxDurationSeconds);
+        let h = Math.trunc(maxDurationSeconds / 60 / 60);
```
JoshuaWilkes: Floor is actually the better fit here, so changing it
```suggestion
        let h = Math.floor(maxDurationSeconds / 60 / 60);
```

JoshuaWilkes:
@jordiup Could you please update the docs with new screenshots for making a request and adding an access rule 
https://docs.commonfate.io/granted-approvals/getting-started/acess-provider
https://docs.commonfate.io/granted-approvals/getting-started/requesting-access

jordiup:
Comments should now be addressed, can be merged once happy :)

JoshuaWilkes (approved):
Thanks for you work on this @jordiup a challenging feature in the end.
I don't expect this will be the final version of this 😆

PR #57, opened 2022-07-25
fix manual revoke - remove second revoke event call

meyerjrr:
The manual revoke was checking state function input for its status. This status will aways be PENDING due to how the logic is now, causing the revoke to never run.
There was also an issue where the revoke event was being run twice.

### DIFF
--- a/accesshandler/pkg/runtime/lambda/revoke_grant.go
+++ b/accesshandler/pkg/runtime/lambda/revoke_grant.go
@@ -8,7 +8,6 @@ import (
 	"github.com/aws/aws-sdk-go-v2/aws"
 	aws_config "github.com/aws/aws-sdk-go-v2/config"
 	"github.com/common-fate/granted-approvals/accesshandler/pkg/config"
-	"github.com/common-fate/granted-approvals/pkg/gevent"
 
 	"github.com/aws/aws-sdk-go-v2/service/sfn"
 	"github.com/common-fate/apikit/logger"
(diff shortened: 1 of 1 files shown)


accesshandler/pkg/runtime/lambda/revoke_grant.go
```diff
@@ -74,15 +73,6 @@ func (r *Runtime) RevokeGrant(ctx context.Context, grantID string, revoker strin
 	//update the grant status
 	grant.Status = types.REVOKED
 
-	eventsBus, err := gevent.NewSender(ctx, gevent.SenderOpts{EventBusARN: r.EventBusArn})
```
JoshuaWilkes, line 77: Where else is this event emitted?

JoshuaWilkes: I see, in the grant service

accesshandler/pkg/runtime/lambda/revoke_grant.go
```diff
@@ -57,11 +56,11 @@ func (r *Runtime) RevokeGrant(ctx context.Context, grantID string, revoker strin
 		return nil, err
 	}
 
-	if grant.Status == types.ACTIVE {
```
JoshuaWilkes, line 60: I'm curious about this.
From my understanding, the revoke endpoint handles both cancelling, 
cancelling is when access is not provisioned in the target but a stepfunction is running, the grant is in the pending state.

A grant in the active state means there is a stepfunction and access is provisioned in the target.
So if revoking is failing because of this if statement, its more likely to me that the grant is never becoming Active rather than this being a bug

meyerjrr: Thats what I thought too. But  since we don't have the concept of a GRANT in the database anymore, `grant` in this case is just the input to the state function. Which doesn't change along the lifecycle (I think) so would always be PENDING.
I guess the error here was that it was looking up the state of the grant from the wrong place?

JoshuaWilkes: image

PR #61, opened 2022-07-27
Improve cdk consistency for inputs and outputs

JoshuaWilkes:
This small improvement ensures that our go type for outputs matches our ts type and reduces some boilerplate in the CDK repo


jordiup (approved):
Looks good Josh

PR #62, opened 2022-07-27
Azure ad provider

meyerjrr:


meyerjrr:
Update azure sso setup to test credentials in cli
Fix Bug causing manual revoke to not revoke access
Fix bug causing providers to return empty json

accesshandler/pkg/providers/azure/ad/ad.go
```diff
@@ -0,0 +1,49 @@
...
+	"github.com/common-fate/granted-approvals/accesshandler/pkg/genv"
+	"github.com/common-fate/granted-approvals/pkg/deploy"
+	"github.com/invopop/jsonschema"
+	"go.uber.org/zap"
+)
+
+const MSGraphBaseURL = "https://graph.microsoft.com/v1.0"
+const ADAuthorityHost = "https://login.microsoftonline.com"
+
+type Provider struct {
```
JoshuaWilkes, line 15: All these properties should be un-exported "lowercase" to help prevent unintentional logging of credentials

accesshandler/pkg/providers/azure/ad/ad.go
```diff
@@ -0,0 +1,49 @@
...
+	Client       AzureClient
+	TenantID     string `yaml:"tenantID"`
```
JoshuaWilkes: Remove all the yaml struct tags

accesshandler/pkg/providers/azure/ad/fixtures/generator.go
```diff
@@ -0,0 +1,103 @@
...
+)
+
+type Fixtures struct {
+	User    string
+	GroupID string
+}
+
+type Generator struct {
+	client       ad.AzureClient
+	tenantID     string `yaml:"tenantID"`
```
JoshuaWilkes: remove yaml struct tags

accesshandler/pkg/runtime/lambda/revoke_grant.go
```diff
@@ -57,32 +56,34 @@ func (r *Runtime) RevokeGrant(ctx context.Context, grantID string, revoker strin
 		return nil, err
 	}
 
-	if grant.Status == types.ACTIVE {
-		err = prov.Provider.Revoke(ctx, string(grant.Subject), args)
-		if err != nil {
-			return nil, err
-		}
+	//do we need to check for grant status here?
```
JoshuaWilkes: Looks to me that this check is in the wrong order.
You want to check the status of the workflow to determine if you need to run revoke or if we only need to "cancel" the existing workflow.

You need to change this around.

JoshuaWilkes (changes requested):

JoshuaWilkes:
UAT
These may be better rectified seperately

- [x] after revoking, the audit log entry shows up as status changed from revoked to revoked
image

- [ ] after cancelling a request, the status shows up as revoked, we should reconsider how this shows to a user, as the access is cancelled before becoming active, not really revoked
image

JoshuaWilkes (approved):

PR #64, opened 2022-07-28
added basic auth error page and update auth logic to show on failed l…

meyerjrr:
If the logging in user is not found in Granteds database previously would cause a infinite loop in the frontend.
This fix will display a helpful error page if this occurs

### DIFF
--- a/web/src/utils/context/userContext.tsx
+++ b/web/src/utils/context/userContext.tsx
@@ -41,6 +41,7 @@ const UserProvider: React.FC<Props> = ({ children }) => {
     // await Auth.currentSession();
 
     const me = await getMe();
+    console.log(me);
     if (me != null) {
       setUser(me.user);
       setIsAdmin(me.isAdmin);
@@ -57,8 +58,8 @@ const UserProvider: React.FC<Props> = ({ children }) => {
       case "cognitoHostedUI":
       case "signOut":
         setUser(undefined);
-        setLoading(false);
-        setamplifyLoggedIn(false);
+        //setLoading(false);
+        //setamplifyLoggedIn(false);
         break;
       case "signIn_failure":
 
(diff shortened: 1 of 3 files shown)


web/src/utils/context/userContext.tsx
```diff
@@ -37,33 +40,36 @@ const UserProvider: React.FC<Props> = ({ children }) => {
 
     // await Auth.currentSession();
 
-    const me = await getMe();
+    const me = await getMe().then((user) => {
```
JoshuaWilkes: Why the change here?
this effectively means that me is never defined?

Using await + .then doesn't make sense

web/src/utils/context/userContext.tsx
```diff
@@ -37,33 +40,36 @@ const UserProvider: React.FC<Props> = ({ children }) => {
...
     switch (event) {
       case "signIn":
+        setamplifyLoggedIn(true);
+        setamplifyUser(data);
       case "cognitoHostedUI":
-        await getUser().then(() => {
+        await getUser().then((out) => {
           if (user !== undefined) {
-            setLoading(false);
+            //setLoading(false );
```
JoshuaWilkes: is this commented out?
Seems like it would still be needed if the call to getMe fails?

JoshuaWilkes (changes requested):
Some comments to clarify

web/src/utils/context/userContext.tsx
meyerjrr: Good catch, was testing a race condition. Reverted it

web/src/utils/context/userContext.tsx
meyerjrr: It called the same thing in the useEffect, so was effectively not needed. Can just be deleted I think

JoshuaWilkes: ok

web/src/utils/context/userContext.tsx
```diff
@@ -38,32 +41,28 @@ const UserProvider: React.FC<Props> = ({ children }) => {
     // await Auth.currentSession();
 
     const me = await getMe();
-    console.debug({ msg: "getMe response", me });
+    console.log(me);
```
JoshuaWilkes:
```suggestion
```

JoshuaWilkes (changes requested):
Looks like tests failing in a strange way, unrelated to this PR. Did we merge something in that caused this?

JoshuaWilkes (approved):

PR #67, opened 2022-07-29
Refactor genv

JoshuaWilkes:


pkg/gconfig/gconfig.go
```diff
@@ -0,0 +1,232 @@
...
+	return "*****"
+}
+
+// MarshalJSON returns a redacted value bytes for this secret
+func (s SecretConfigValue) MarshalJSON() ([]byte, error) {
+	return json.Marshal(s.String())
+}
+
+// ConfigValue value implements the Valuer interface
+type ConfigValue string
```
chrnorm: `ConfigValue` obscures the fact that it's a string under the hood. What about `StringValue`?

```suggestion
type StringValue string
```

pkg/gconfig/gconfig.go
```diff
@@ -0,0 +1,232 @@
...
+	if s.value == nil {
+		return ""
+	}
+	return s.value.String()
+}
+
+// SecretConfigValue value implements the Valuer interface, it should be used for secrets in configuration structs.
+//
+// It is configured to automatically redact the secret for common logging usecases like Zap, fmt.Println and json.Marshal
+type SecretConfigValue string
```
chrnorm: same thinking as previous comment:

```suggestion
type SecretStringValue string
```

pkg/gconfig/json_loader.go
```diff
@@ -0,0 +1,94 @@
...
+	"sync"
+
+	"github.com/aws/aws-sdk-go-v2/service/ssm"
+	"github.com/common-fate/granted-approvals/pkg/cfaws"
+	"github.com/pkg/errors"
+	"golang.org/x/sync/errgroup"
+)
+
+type SecretGetter interface {
+	Get(ctx context.Context, path string) (string, error)
```
chrnorm: `Get` is quite similar to `Load` (the method used in `JSONLoader`) which I feel could be confusing as they have different use cases - loading all config vs retrieving individual secrets

```suggestion
	GetSecret(ctx context.Context, path string) (string, error)
```

pkg/gconfig/gconfig.go
```diff
@@ -0,0 +1,232 @@
...
+	}
+	return nil
+}
+
+// Dump renders a map[string]string where the values are mapped in different ways based on the provided dumper
+//
+// use SafeDumper to get all values with secrets redacted
+//
+// SSMDumper first pushes any updated secrets to ssm then returns the ssm paths to the secrets
+func (c Config) Dump(ctx context.Context, dumper Dumper) (map[string]string, error) {
```
chrnorm, line 57: Do you think we need the `dumper` interface? I'm wondering if it might be simpler to just have different structs with a `Dump` method e.g.

```go
var c Config
s := SSMDumper{}
s.Dump(ctx, c)
```

rather than the current approach:
```go
var c Config
s := SSMDumper{}
c.Dump(ctx, s)
```

My only criticism of the current approach is that it feels a bit less intuitive as you'd expect the SSMDumper to 'do' the dumping, (`s.Dump()`) rather than the config to do the dumping.

We might also have this problem with the `config.Load()` method which I originally wrote!

chrnorm (commented):
Looks really good, the secrets handling in particular is a big safety improvement over my initial attempt. My comments are all nitpick/naming considerations only

pkg/gconfig/gconfig.go
JoshuaWilkes: My original reasoning for this was a need to avoid a naming collision, but since then I renamed the underlying struct from Value to Field so I will update the Creator functions from String() and SecretString() to be Field() and SecretField()
Though I think I could event simplify it further by just having Field(SecretStringValue || StringValue) and infer whether its a secret from the value type

JoshuaWilkes: Though that actually reduces readability so I will stick to SecretField() and Field()

pkg/gconfig/gconfig.go
JoshuaWilkes: I think I will just make the safeDump and SSMDump methods on Config directly

JoshuaWilkes: interesting point. I think its important for "easy to use, hard to misuse" that if you have a config, its obvious how you get all the values out.

Our specific use-case requires SSM to be involved. 

As a developer using this library i think, "i have a Config, how do I get the values out"
I use autocomplete to list the methods available then I read the description of Dump.
If we removed dump there would be no autocomplete :(
I propose we leave it in for now and can factor it out later if its not clear enough.

meyerjrr (approved):

PR #69, opened 2022-08-01
CF-85: Slack, Reviewer and Request Time

jordiup:


pkg/gevent/request_events.go
```diff
@@ -23,6 +23,7 @@ func (RequestCreated) EventType() string {
 // user's request is approved.
 type RequestApproved struct {
 	Request access.Request `json:"request"`
+	Review  access.Review  `json:"review"`
```
JoshuaWilkes: I propose changing these to just include the reviewer ID, the reason being that the Reviewer struct is almost entirely a clone of the Request, so you will just be duplicating the data. 

Given that the event type indicate the review response already, e.g approved or declined we don't need to include this extra information in the event body. Same goes for the other events changes below
```suggestion
	ReviewerID string `json:"reviewerId"`
```

pkg/notifiers/slack/handle_request.go
```diff
@@ -189,7 +220,20 @@ func (n *Notifier) HandleRequestEvent(ctx context.Context, log *zap.SugaredLogge
 	return nil
 }
 
-func (n *Notifier) UpdateSlackMessage(ctx context.Context, slackClient *slack.Client, log *zap.SugaredLogger, rev access.Reviewer, req access.Request, rule rule.AccessRule, dbRequestor *identity.User) error {
+type UpdateSlackMessageOpts struct {
+	Ctx           context.Context
```
JoshuaWilkes: In Go, it is not idiomatic to put context inside a struct, instead, it should always be passed as the first argument in a function when needed.

I would also preference only including "data" in the opts for something like this, and keeping slackSlient and log as parameters in the function call
```
UpdateSlackMessage(ctx context.Context, slackClient *slack.Client, log *zap.SugaredLogger, opts UpdateSlackMessageOpts)
```

pkg/notifiers/slack/handle_request.go
```diff
@@ -189,7 +220,20 @@ func (n *Notifier) HandleRequestEvent(ctx context.Context, log *zap.SugaredLogge
...
+	Review        access.Reviewer
+	Request       access.Request
+	RequestReview *access.Review
+	Rule          rule.AccessRule
+	DbRequestor   *identity.User
+}
+
+func (n *Notifier) UpdateSlackMessage(opts UpdateSlackMessageOpts) error {
+
+	ctx, slackClient, log, rev, req, reqRev, rule, dbRequestor := opts.Ctx, opts.SlackClient, opts.Log, opts.Review, opts.Request, opts.RequestReview, opts.Rule, opts.DbRequestor
```
JoshuaWilkes: Where possible, avoiding this type of "destructuring" is preferred
In this case it looks like you have added this to avoid updating the existing function.

My preference would be updating the function to use opts.RequestReview.something etc rather than reqRev as its harder to read where the values come from

JoshuaWilkes (changes requested):
I have requested some stylistic changes to this PR

pkg/notifiers/slack/handle_request.go
JoshuaWilkes: Some content from [ardanlabs]( https://www.ardanlabs.com/blog/2019/09/context-package-semantics-in-go.html#:~:text=It's%20an%20idiom%20in%20Go,pointer%20semantics%20should%20be%20used.&text=Every%20function%20that%20accepts%20a,copy%20of%20the%20interface%20value.) on context

pkg/notifiers/slack/handle_request.go
```diff
@@ -189,44 +208,59 @@ func (n *Notifier) HandleRequestEvent(ctx context.Context, log *zap.SugaredLogge
...
+	DbRequestor       *identity.User
+}
+
+func (n *Notifier) UpdateSlackMessage(ctx context.Context, slackClient *slack.Client, log *zap.SugaredLogger, opts UpdateSlackMessageOpts) error {
 
 	// Skip if requestor == reviewer
-	if rev.ReviewerID == req.RequestedBy {
-		return nil
-	}
+	// if opts.Review.ReviewerID == opts.Request.RequestedBy  {
```
JoshuaWilkes: @jordiup is this meant to b commented out?
if its not longer required it should be removed

JoshuaWilkes (changes requested):
Just one comment otherwise looks good

pkg/notifiers/slack/handle_request.go
```diff
@@ -189,44 +208,59 @@ func (n *Notifier) HandleRequestEvent(ctx context.Context, log *zap.SugaredLogge
...
+	// 	return nil
+	// }
```
jordiup:
```suggestion
```

pkg/notifiers/slack/handle_request.go
jordiup: Cheers Josh, correct that it can be removed. I've just pushed a commit :)

JoshuaWilkes (approved):

PR #70, opened 2022-08-02
Add docs links to cli

meyerjrr:
Made linking to docs easier
More clear at which point in the setup process where to get help from the docs


cmd/gdeploy/utils/linker.go
```diff
@@ -0,0 +1,14 @@
+package linker
+
+// Helper package to clean up the links to the docs in the gdeploy cli
+type Linker struct {
```
chrnorm: We can remove this and use `url.URL` which is built in to the Go standard library for joining paths: 
```go
u := url.URL{
	Scheme: "https",
	Host: "docs.commonfate.io",
}
u.Path = "granted-approvals/sso/overview"
fmt.Printf("%s", u.String())
```

Separately - I usually request changes on using packages like `utils` as it's good to use a more specific package name in Go :) (but in this case we can remove the `utils` folder altogether)

chrnorm (changes requested):

cmd/gdeploy/commands/sso/sso.go
```diff
@@ -56,6 +57,10 @@ var configureCommand = cli.Command{
 		azureConfigured := dc.Identity != nil && dc.Identity.Azure != nil
 		isCurrentIDP := dc.Deployment.Parameters.IdentityProviderType == strings.ToUpper(idpType)
 		update := true
+
+		clio.Warn("Don't know where to find a SSO credential? Best place to find out would be our docs!")
```
chrnorm:
```suggestion
		clio.Warn("Don't know where to find an SSO credential? Best place to find out would be our docs!")
```

chrnorm (dismissed):

cmd/gdeploy/utils/linker.go
meyerjrr: Perfect, didn't know about this functionality in `url.URL` thanks

JoshuaWilkes (approved):

PR #71, opened 2022-08-02
Jordi/cf 17 write frontend tests for requestapproval

jordiup:


web/src/pages/requests/index.tsx
```diff
@@ -348,6 +350,8 @@ const UserAccessCard: React.FC<{
             flexDir="column"
             key={req.id}
             pos="relative"
+            // id={"req_"+ index}
+            id={"req_" + req.reason}
```
chrnorm: Can we use `data-testid` as an attribute rather than `id` here? To prevent potential ID collisions in HTML

JoshuaWilkes (approved):

PR #75, opened 2022-08-03
refactor to be testable with maps

meyerjrr:


JoshuaWilkes (changes requested):
It looks like this will work
Though I have actually solved this a diferent way in my own PR

I just added `sort.Strings(opts)` before the return in CLIOptions which ensures that we always have consistent order.

There was actually a seperate UX issue from this where the cli would show the options in a different order each time.

Can you update this PR to just have the sort code added and don't split the function up as it will cause a conflic with my branch otherwise.

Maybe just open a new PR for this

JoshuaWilkes (approved):

PR #113, opened 2022-08-11
Access Rule Error Message

### LINKED ISSUE #97: When creating an access rule, an error messages shows when only a single approver is selected, not a group
DE = the error message should revalidate when either of the approval fields are touched
UDE = error message shows when the input is valid

jordiup:
Recording 2022-08-11 at 15 44 15
This should cover our desired effects :)

### DIFF
--- a/web/src/components/forms/access-rule/steps/Approval.tsx
+++ b/web/src/components/forms/access-rule/steps/Approval.tsx
@@ -11,7 +11,6 @@ import {
   Wrap,
   WrapItem,
 } from "@chakra-ui/react";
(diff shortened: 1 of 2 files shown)


web/src/components/forms/access-rule/steps/Approval.tsx
```diff
@@ -11,6 +11,7 @@ import {
   Wrap,
   WrapItem,
 } from "@chakra-ui/react";
+import { triggerAsyncId } from "async_hooks";
```
chrnorm:
```suggestion
```

chrnorm (approved):

Issue #118, opened 2022-08-12
AWS SSO provider should only list Permission Sets defined in the account it runs in. e.g the Delegated admin account.


JoshuaWilkes:
Done = when creating an access rule for sso, only permissions sets in the delegated admin account are shows in the options.

The UDE is that when running in a delegated admin account, granted cannot assign users to permission sets in the management account. Though the are still shown in the UI leading to issues for our users where they create an access rule which fails to grant every time.

PR #119, opened 2022-08-12
Add tag check to permission set list method

JoshuaWilkes:
This PR fixes a UX issue by introducing a resource tag check for th epermission sets.
In AWS there seems to be no way via the API to determine if a permission set belongs to the delegated account or not. Therefor we found the most suitable way to limit the permission set results is to have admins add a specific tag to the permission sets when they create them.

We will add this into our docs for SSO setup.

I have chosen a generic tag name `commonfate.io/permission-set` We should probably put some more though into this tag so that we can use it consistently for other resources in the future should we need to.

For example, permission sets generated by the access handler or other resources managed by granted

### DIFF
--- a/accesshandler/pkg/providers/aws/sso/options.go
+++ b/accesshandler/pkg/providers/aws/sso/options.go
@@ -95,7 +95,7 @@ func (p *Provider) checkPermissionSetIsTagged(ctx context.Context, permissionSet
 		nextToken = tags.NextToken
 		hasMore = nextToken != nil
 		for _, tag := range tags.Tags {
-			if aws.ToString(tag.Key) == "commonfate.io/permission-set" {
+			if aws.ToString(tag.Key) == "commonfate.io/managed-by-granted" {
 				return true, nil
 			}
 		}
--- a/accesshandler/pkg/providers/aws/sso/options.go
+++ b/accesshandler/pkg/providers/aws/sso/options.go
@@ -20,26 +20,40 @@ func (p *Provider) Options(ctx context.Context, arg string) ([]types.Option, err
 		opts := []types.Option{}
 		hasMore := true
 		var nextToken *string
+
 		for hasMore {
 			o, err := p.client.ListPermissionSets(ctx, &ssoadmin.ListPermissionSetsInput{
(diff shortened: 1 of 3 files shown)


JoshuaWilkes:
Fixes #118

accesshandler/pkg/providers/aws/sso/options.go
```diff
@@ -66,3 +80,25 @@ func (p *Provider) Options(ctx context.Context, arg string) ([]types.Option, err
...
+			InstanceArn: aws.String(p.instanceARN.Get()),
+			ResourceArn: aws.String(permissionSetARN),
+		})
+		if err != nil {
+			return false, err
+		}
+		nextToken = tags.NextToken
+		hasMore = nextToken != nil
+		for _, tag := range tags.Tags {
+			if aws.ToString(tag.Key) == "commonfate.io/permission-set" {
```
chrnorm: We can use a more generic tag to indicate the resource is managed/accessible by granted

```suggestion
			if aws.ToString(tag.Key) == "commonfate.io/managed-by-granted" {
```

JoshuaWilkes: Related to this, for EKS provider granted also creates permission sets, we shoudl probably tag those too, but we don't want them to show up in the UI for the SSO provider.

maybe for these EKS ones we use `commonfate.io/created-by-granted` 

I also think that I should be updating the SSO provider validation check to ensure that only permissions sets with this tag can be granted as part of this PR

chrnorm (dismissed):
> I also think that I should be updating the SSO provider validation check to ensure that only permissions sets with this tag can be granted as part of this PR

@JoshuaWilkes this is a great idea - I've marked this as a Request Changes on this PR

accesshandler/pkg/providers/aws/sso/options.go
JoshuaWilkes: @chrnorm I have added this, but I realised that the Validater interface is only used in the integration test, It's not clear to me what your original intention was for this.

I will leave it out of scope for this ticket

PR #123, opened 2022-08-12
Access Provider Select Fields

### LINKED ISSUE #98: Access Provider select fields should be sorted alphabetically and searchable
When creating an Access Rule it's difficult to find a particular role or account if many exist.

Currently, we use a native select drop-down UI, as shown in the below screenshot:

Image

Done = options in the select field are sorted alphabetically
Done = options can be searched

The UI should be updated to work similar to the group selector:

Image

jordiup:
* Improvement to use keyboard entry style inpu
Recording 2022-08-12 at 14 40 48


chrnorm (approved):
Looks great @jordiup!

PR #131, opened 2022-08-15
hotfix access handler error handling

jordiup:
image

This is a small hotfix to extend the null check. Our OpenAPI Client doesn't seem to include `null` as a response value for `/api/v1/requests/${requestId}/access-instructions` however as can be seen in the above screenshot it is necessary to handle this value


JoshuaWilkes:
- API schema for access handler api instruction does not match teh implementation, returns 404 but spec says 400
- de = when a provider does not exist, the access handler api returns a 404 with an error message. This should be forwarded through to the FE, the frontend should handle a 404 error correctly. a 404 would indicate that a provider has been deleted. We can probably fix this patially by only showing instructions for active requests. 

DE = in the case where a request is active but a provider is not found, it could mean that the access rule has not been archived, and someone requested it. in which case the request would never be successful, in any case, thr frontend should just hide the instructions if there is an error response.

jordiup:
> * API schema for access handler api instruction does not match teh implementation, returns 404 but spec says 400
> * de = when a provider does not exist, the access handler api returns a 404 with an error message. This should be forwarded through to the FE, the frontend should handle a 404 error correctly. a 404 would indicate that a provider has been deleted. We can probably fix this patially by only showing instructions for active requests.
> 
> DE = in the case where a request is active but a provider is not found, it could mean that the access rule has not been archived, and someone requested it. in which case the request would never be successful, in any case, thr frontend should just hide the instructions if there is an error response.

@JoshuaWilkes think this should meet these DEs now?

JoshuaWilkes:
@jordiup You will need to update the accesshandler api spec or the api implementation so that the error response in the spec matches the implementation.

If the two don't match, then the api client will fail to unmarshal the response correctly into JSON400, can you update this please?

jordiup:
> @jordiup You will need to update the accesshandler api spec or the api implementation so that the error response in the spec matches the implementation.
> 
> If the two don't match, then the api client will fail to unmarshal the response correctly into JSON400, can you update this please?

Cheers Josh, have finally seen what you're saying. AccessHandler spec amended with corresponding handling in API :) Cheers

pkg/api/request.go
```diff
@@ -324,7 +324,15 @@ func (a *API) GetAccessInstructions(w http.ResponseWriter, r *http.Request, requ
...
 
-	apio.JSON(ctx, w, res.JSON200, http.StatusOK)
+	switch res.StatusCode() {
+	case http.StatusOK:
+		apio.JSON(ctx, w, res.JSON200, http.StatusOK)
+	case http.StatusNotFound:
+		apio.JSON(ctx, w, res.JSON404.Error, res.StatusCode())
+	case http.StatusBadRequest:
+		apio.JSON(ctx, w, res.JSON400.Error, res.StatusCode())
+	}
```
JoshuaWilkes, line 336: We should have a final default error response here just to be sure

JoshuaWilkes (changes requested):
One final addition then I think its good to go

pkg/api/request.go
jordiup: Cheers Josh, should be covered now :)

JoshuaWilkes (approved):
Looks good!

PR #133, opened 2022-08-15
Improve experience for selecting admin group

### LINKED ISSUE #115: Improve experience for selecting a granted administrator group ID for SSO
Once we have SSO config we can list groups via the API and use a survey.Select prompt to list the options, 

Ask if they want to update if its already set -> ask if they want to list options or enter an ID - prompt with options or accept ID
Then finally verify that the group exists before setting the config value and saving

The goal is to make it easier for an admin when they run gdeploy sso configure
At the end of this command flow, when they have connected their azure AD or okta SSO, the CLI prompts then to set the administrator Group ID.
This is what determines if a user is an admin.
Once SSO is configured, use the listGroups method of the identityprovider and display the options using a survey.Select prompt if the users chooses to else, allow them to enter an id directly.
DE = before prompting to list all groups, ask the user if they want to enter an id or select from a list of their groups. so you can just enter it if you already have it
DE = if they want to list, they should be able to search the groups by name. if possible, show the description too if we have that information?
DE = if they enter a group id directly, validate that the group exists. You can implement this as a validator on the survey.Input prompt so they get promopted to re enter the id if it does not exist.

jordiup:
PR for #115 :)

### DIFF
--- a/cmd/gdeploy/commands/sso/sso.go
+++ b/cmd/gdeploy/commands/sso/sso.go
@@ -3,11 +3,13 @@ package sso
 import (
 	"fmt"
 	"os"
+	"sort"
 
 	"github.com/AlecAivazis/survey/v2"
 	"github.com/common-fate/granted-approvals/pkg/clio"
 	"github.com/common-fate/granted-approvals/pkg/deploy"
 	"github.com/common-fate/granted-approvals/pkg/gconfig"
+	"github.com/common-fate/granted-approvals/pkg/identity"
 	"github.com/common-fate/granted-approvals/pkg/identity/identitysync"
 
 	"github.com/urfave/cli/v2"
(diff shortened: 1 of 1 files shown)


cmd/gdeploy/commands/sso/sso.go
```diff
@@ -176,12 +179,71 @@ var configureCommand = cli.Command{
 Create a group called 'Granted Administrators' in your identity provider and copy the group's ID.
 Users in this group will be able to manage Access Rules.
 `)
-		adminGroupPrompt := &survey.Input{Message: "The ID of the Granted Administrators group in your identity provider:", Default: dc.Deployment.Parameters.AdministratorGroupID}
-		err = survey.AskOne(adminGroupPrompt, &dc.Deployment.Parameters.AdministratorGroupID)
+
+		// Prompt: Do you want to enter an group id, or select frokm a list of groups?
+		surveyQ := []string{"Select from a list of groups", "Enter group id"}
```
chrnorm: Instead of giving an option here, can we always prompt the user to select from a list of groups? I think this could add to confusion for first-time setup as users won't be sure whether they should choose from a list or enter an ID. Additionally, if users want to manually enter an ID they can just edit the `granted-deployment.yml` file by hand.

chrnorm (dismissed):
Looks good @jordiup! One requested change from me

cmd/gdeploy/commands/sso/sso.go
jordiup: Cheers @chrnorm, think this should be good now. I agree that it makes for a simpler UX (eliminating choice)

cmd/gdeploy/commands/sso/sso.go
```diff
@@ -176,12 +179,45 @@ var configureCommand = cli.Command{
...
 `)
-		adminGroupPrompt := &survey.Input{Message: "The ID of the Granted Administrators group in your identity provider:", Default: dc.Deployment.Parameters.AdministratorGroupID}
-		err = survey.AskOne(adminGroupPrompt, &dc.Deployment.Parameters.AdministratorGroupID)
+
+		err = idp.IdentityProvider.Init(ctx)
+		if err != nil {
+			return err
+		}
+
+		err = deploy.RunConfigTest(ctx, idp.IdentityProvider)
```
JoshuaWilkes: We should only run the test if `update == false`
The UDE = I have updated my config and I see a test passed prompt, then straight away I see tests passed again.

Alternatively we can actually just omit the test here all together because I think it is unessecary.

```suggestion
		
```

cmd/gdeploy/commands/sso/sso.go
```diff
@@ -176,12 +179,45 @@ var configureCommand = cli.Command{
...
+		err = survey.AskOne(&survey.Select{
+			Message: "The ID of the Granted Administrators group in your identity provider:",
+			Options: groupNames,
+		}, &chosenKey)
+
+		if err != nil {
+			return err
+		}
+
+		dc.Deployment.Parameters.AdministratorGroupID = groupMap[chosenKey].Name
```
JoshuaWilkes: This needs to be the ID rather than the name
```suggestion
		dc.Deployment.Parameters.AdministratorGroupID = groupMap[chosenKey].ID
```

JoshuaWilkes (changes requested):
I have requested some minor changes, looking good

cmd/gdeploy/commands/sso/sso.go
```diff
@@ -176,12 +179,45 @@ var configureCommand = cli.Command{
...
-		adminGroupPrompt := &survey.Input{Message: "The ID of the Granted Administrators group in your identity provider:", Default: dc.Deployment.Parameters.AdministratorGroupID}
-		err = survey.AskOne(adminGroupPrompt, &dc.Deployment.Parameters.AdministratorGroupID)
+
+		err = idp.IdentityProvider.Init(ctx)
+		if err != nil {
+			return err
+		}
+
+		
+		if err != nil {
```
JoshuaWilkes: Can remove this extra if err block here

JoshuaWilkes (changes requested):
One last tidy up then good to go 🚀

cmd/gdeploy/commands/sso/sso.go
```diff
@@ -176,12 +179,45 @@ var configureCommand = cli.Command{
...
+			return err
+		}
```
JoshuaWilkes:
```suggestion
```

cmd/gdeploy/commands/sso/sso.go
```diff
@@ -176,12 +179,45 @@ var configureCommand = cli.Command{
...
-		err = survey.AskOne(adminGroupPrompt, &dc.Deployment.Parameters.AdministratorGroupID)
+
+		err = idp.IdentityProvider.Init(ctx)
+		if err != nil {
+			return err
+		}
+
+		
+		if err != nil {
+			return err
```
JoshuaWilkes:
```suggestion
```

cmd/gdeploy/commands/sso/sso.go
JoshuaWilkes:
```suggestion
```

JoshuaWilkes (approved):

PR #146, opened 2022-08-16
Pagination Next Page

### LINKED ISSUE #93: Pagination buttons should show the next page if nextToken is returned from the API

jordiup:
PR for #93 
image

Now shows the next page if nextToken is returned from the API
Can demo on deployment :)

### DIFF
--- a/web/src/utils/usePaginatorApi.ts
+++ b/web/src/utils/usePaginatorApi.ts
@@ -56,7 +56,7 @@ export const usePaginatorApi = <T extends (...args: any[]) => any>({
   const [pageOptions, setPageOptions] = useState<number[]>([pageIndex]);
 
   useEffect(() => {
-    if (data && data.next && pageOptions.length == 1) {
+    if (data?.next && pageOptions?.length === 1) {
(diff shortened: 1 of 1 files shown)


web/src/utils/usePaginatorApi.ts
```diff
@@ -55,6 +55,13 @@ export const usePaginatorApi = <T extends (...args: any[]) => any>({
 
   const [pageOptions, setPageOptions] = useState<number[]>([pageIndex]);
 
+  useEffect(() => {
+    if (data && data.next && pageOptions.length == 1) {
```
Eddie023:
```suggestion
    if (data?.next && pageOptions?.length === 1) {
```
Not sure but I think we can use [optional chaining](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html). I think our typescript version supports this. 

Not sure if `pageOptions` returns an empty array by default; if not then it will cause an error when trying to get `length` of `undefined` 

`===` instead of `==`

JoshuaWilkes (approved):

PR #151, opened 2022-08-17
Add infinite scroll paginated list views

### LINKED ISSUE #148: Add Infinite-Scroll Paginated List Views
Inspiration: https://swr.vercel.app/examples/infinite-loading

DE = Infinite loading is added to key list views (that aren't already tables)
DE = Scrolling down appends freshly loaded items to the list

jordiup:
- [x] DE = list response loads normally
- [x] DE = when scrolling to bottom, pagination is triggered
- [x] DE = the paginated response is appended to a *virtual* list (should contain all pages)
- [x] DE = mutate/revalidation queries still update the UI
https://www.loom.com/share/be002205ac834a77b9b11cf5634ab29a


JoshuaWilkes:
In your loom, you say that there is an intersect api which does not work every time as expected. 
Rather than ship it with this change, I think it would be better to just have the load more button and a large query limit.
For the "upcoming requests" it's probably unlikely that you will have more than 10? at any one time. 
Past request is where you will get a build up of requests over time, Its probably unlikely that end users will be scrolling back very far in this list.

Rather, admins will look back at old requests in the requests tables in teh admin UI.

So my opinion on this is to remove the autoload feature/intersect API and just have the load more button which is working well.
Do you think that is reasonable?

We could add intersection at a later date if needed.

JoshuaWilkes:
@jordiup I think you can merge main into this branch now to reduce the review required as I have merged your pagination PR in already

pkg/api/request.go
```diff
@@ -36,14 +36,20 @@ func (a *API) UserListRequestsUpcoming(w http.ResponseWriter, r *http.Request, p
 		RequestEndComparator: storage.GreaterThan,
 		CompareTo:            time.Now(),
 	}
-	_, err := a.DB.Query(ctx, &q, queryOpts...)
+	qR, err := a.DB.Query(ctx, &q, queryOpts...)
```
JoshuaWilkes: Real nitpick here 😆 can this just be qr instead of qR
```suggestion
	qr, err := a.DB.Query(ctx, &q, queryOpts...)
```

JoshuaWilkes (changes requested):
one minor nitpick 😁

pkg/api/request.go
```diff
@@ -57,21 +63,32 @@ func (a *API) UserListRequestsPast(w http.ResponseWriter, r *http.Request, param
...
+
 	// the items in the list will be sorted by the request endtime not requestedAt
 	// is this going to be a problem?
 	q := storage.ListRequestsForUserAndRequestend{
 		UserID:               uid,
 		RequestEndComparator: storage.LessThanEqual,
 		CompareTo:            time.Now(),
 	}
-	_, err := a.DB.Query(ctx, &q)
+	qR, err := a.DB.Query(ctx, &q, queryOpts...)
```
JoshuaWilkes:
```suggestion
	qr, err := a.DB.Query(ctx, &q, queryOpts...)
```

pkg/api/request.go
```diff
@@ -57,21 +63,32 @@ func (a *API) UserListRequestsPast(w http.ResponseWriter, r *http.Request, param
...
 	if err != nil && err != ddb.ErrNoItems {
 		apio.Error(ctx, w, err)
 		return
 	}
 
+	var next *string
+	if qR.NextPage != "" {
```
JoshuaWilkes:
```suggestion
	if qr.NextPage != "" {
```

JoshuaWilkes (approved):

pkg/api/request.go
```diff
@@ -57,21 +63,32 @@ func (a *API) UserListRequestsPast(w http.ResponseWriter, r *http.Request, param
...
-	_, err := a.DB.Query(ctx, &q)
+	qr, err := a.DB.Query(ctx, &q, queryOpts...)
 	if err != nil && err != ddb.ErrNoItems {
 		apio.Error(ctx, w, err)
 		return
 	}
 
+	var next *string
+	if qr.NextPage != "" {
+		next = &qR.NextPage
```
JoshuaWilkes:
```suggestion
		next = &qr.NextPage
```

PR #161, opened 2022-08-18
Verify if gdeploy build version and granted-approval release version matches

Eddie023:
### Description 
1.  If `gdeploy -v` && `release` value in `granted-deployment.yml` is different. Prompt users to update the value inside `granted-deployment.yml` to match `gdeploy -v`. Save the updated config. 
2. If users choose not to update then the cmd will return with an error. 

### Summary of changes 
1.  Create a wrapper beforeFunc `VerifyGDeployCompatibility` 
2.  Add `IsReleaseVersionDifferent` func that will return `(bool, error)`
    a. True -> if version is different 
    b. False -> if everything is okay; if the check can be skipped 
    c. Error -> If something returns err 
3. Add unit tests to verify different conditions.


cmd/gdeploy/main.go
```diff
@@ -268,3 +270,36 @@ func PreventDevUsage() cli.BeforeFunc {
...
+		isValidReleaseNumber, err := regexp.MatchString(`v\d.\d+.\d+`, dc.Deployment.Release)
+		if err != nil {
+			return err
+		}
+
+		if isValidReleaseNumber && build.Version != dc.Deployment.Release {
+			return clio.NewCLIError("Uncompatible gdeploy version and granted-approval version. You need to update gDeploy to the latest version.")
+		}
+
+		// release value are added as URL for UAT. In such case it should skip this check.
```
JoshuaWilkes: This block needs to come above the isValidReleaseNumber check otherwise url based releases will always fail

cmd/gdeploy/main.go
```diff
@@ -268,3 +270,36 @@ func PreventDevUsage() cli.BeforeFunc {
...
+		}
+
+		// skip compatibility check for dev deployments.
+		if dc.Deployment.Dev != nil && *dc.Deployment.Dev {
+			return nil
+		}
+
+		isValidReleaseNumber, err := regexp.MatchString(`v\d.\d+.\d+`, dc.Deployment.Release)
+		if err != nil {
+			return err
```
JoshuaWilkes, line 288: I think we can omit this check entirely, we only need to know if the release number matches the build.Version and if not inform teh user with an error message showing what they had vs what was expected

cmd/gdeploy/main.go
```diff
@@ -58,8 +60,8 @@ func main() {
 			WithBeforeFuncs(&groups.GroupsCommand, RequireDeploymentConfig(), RequireAWSCredentials()),
 			WithBeforeFuncs(&logs.Command, RequireDeploymentConfig(), RequireAWSCredentials()),
 			WithBeforeFuncs(&commands.StatusCommand, RequireDeploymentConfig(), RequireAWSCredentials()),
-			WithBeforeFuncs(&commands.CreateCommand, RequireDeploymentConfig(), PreventDevUsage(), RequireAWSCredentials(), RequireCleanGitWorktree()),
-			WithBeforeFuncs(&commands.UpdateCommand, RequireDeploymentConfig(), PreventDevUsage(), RequireAWSCredentials(), RequireCleanGitWorktree()),
+			WithBeforeFuncs(&commands.CreateCommand, RequireDeploymentConfig(), PreventDevUsage(), VerifyGDeployCompatibility(), RequireAWSCredentials(), RequireCleanGitWorktree()),
```
JoshuaWilkes, line 64: You should add this check to all commends except for Init

cmd/gdeploy/main.go
```diff
@@ -268,3 +270,36 @@ func PreventDevUsage() cli.BeforeFunc {
 		return nil
 	}
 }
+
+// Check whether gdeploy version matches with the granted-approval.
+func VerifyGDeployCompatibility() cli.BeforeFunc {
```
JoshuaWilkes, line 278: it would be good if we could unit test this

JoshuaWilkes (changes requested):

cmd/gdeploy/main.go
```diff
@@ -54,21 +57,21 @@ func main() {
...
-			WithBeforeFuncs(&commands.StatusCommand, RequireDeploymentConfig(), RequireAWSCredentials()),
-			WithBeforeFuncs(&commands.CreateCommand, RequireDeploymentConfig(), PreventDevUsage(), RequireAWSCredentials(), RequireCleanGitWorktree()),
-			WithBeforeFuncs(&commands.UpdateCommand, RequireDeploymentConfig(), PreventDevUsage(), RequireAWSCredentials(), RequireCleanGitWorktree()),
-			WithBeforeFuncs(&sso.SSOCommand, RequireDeploymentConfig(), RequireAWSCredentials()),
-			WithBeforeFuncs(&backup.Command, RequireDeploymentConfig(), PreventDevUsage(), RequireAWSCredentials()),
-			WithBeforeFuncs(&restore.Command, RequireDeploymentConfig(), PreventDevUsage(), RequireAWSCredentials()),
-			WithBeforeFuncs(&provider.Command, RequireDeploymentConfig(), RequireAWSCredentials()),
-			WithBeforeFuncs(&notifications.Command, RequireDeploymentConfig(), RequireAWSCredentials()),
-			WithBeforeFuncs(&dashboard.Command, RequireDeploymentConfig(), RequireAWSCredentials()),
+			WithBeforeFuncs(&users.UsersCommand, RequireDeploymentConfig(), RequireAWSCredentials(), VerifyGDeployCompatibility()),
```
JoshuaWilkes: @Eddie023 I think that the version check should have precedence over the AWS credentials check, this way, the user gets the most important error first.

Could you swap this around for all the commands?

```suggestion
			WithBeforeFuncs(&users.UsersCommand, RequireDeploymentConfig(),  VerifyGDeployCompatibility(),RequireAWSCredentials()),
```

JoshuaWilkes (changes requested):
Requesting a change to the order of the before funcs, otherwise this looks good!

JoshuaWilkes (approved):

PR #171, opened 2022-08-23
gdeploy: allow "-y" to be used as confirmation

chrnorm:
adds a `-y` alias to the `--confirm` argument.

### DIFF
--- a/cmd/gdeploy/commands/create.go
+++ b/cmd/gdeploy/commands/create.go
@@ -13,7 +13,7 @@ var CreateCommand = cli.Command{
 	Usage:       "Create a new Granted Approvals deployment by deploying CloudFormation",
 	Description: "Create a new Granted Approvals deployment based on a deployment configuration file (granted-deployment.yml by default). Deploys resources to AWS using CloudFormation.",
 	Flags: []cli.Flag{
-		&cli.BoolFlag{Name: "confirm", Usage: "If provided, will automatically deploy without asking for confirmation"},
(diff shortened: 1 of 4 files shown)


cmd/gdeploy/commands/create.go
```diff
@@ -13,7 +13,7 @@ var CreateCommand = cli.Command{
 	Usage:       "Create a new Granted Approvals deployment by deploying CloudFormation",
 	Description: "Create a new Granted Approvals deployment based on a deployment configuration file (granted-deployment.yml by default). Deploys resources to AWS using CloudFormation.",
 	Flags: []cli.Flag{
-		&cli.BoolFlag{Name: "confirm", Usage: "If provided, will automatically deploy without asking for confirmation"},
+		&cli.BoolFlag{Name: "confirm", Aliases: []string{"y"}, Usage: "If provided, will automatically deploy without asking for confirmation"},
```
Eddie023, line 16: @chrnorm  Quick question on this. So, in my last PR I have added a prompt that will ask user if they like to update their `granted-deployment.yml` version if there is version mismatch. Should I make some changes on based on this PR? I think if there is a version mismatch `gdepoy create -y` will still prompt users with a question. 

cc: @JoshuaWilkes

JoshuaWilkes: @Eddie023 in the case of your PR

if you added an override, (there may already be one)
it would be `gdeploy -confirm create -y`

JoshuaWilkes (approved):

PR #173, opened 2022-08-23
implement provider argument options caching

### LINKED ISSUE #172: AWS SSO options lambda times out after 20 seconds if SSO instance is in a different region to the Granted deployment
If the SSO instance is in `ap-southeast-2` and Granted is deployed to `us-west-2`, the Lambda times out after 20 seconds before the `/api/v1/admin/providers/aws-sso/args/permissionSetArn/options` can complete.

Looking at the code in `accesshandler/pkg/providers/aws/sso/options.go`, we're not making use of any concurrency and we have some nested for loops which make API calls describe each permission set and list all of their tags. This means that the more permission sets a team has, the longer the endpoint will take.

We should update this function with the following approach:

1. call the `ListPermissionSets` API to find all of the Permission Sets (looping through pages if there are any)
2. launch a goroutine for each permission set
3. inside each goroutine, call the `DescribePermissionSet` and `checkPermissionSetIsManagedByGranted` functions
4. inside each goroutine, push the results to the `opts` `[]types.Option{}` struct.

A longer term solution is to cache the options in our database, but this will introduce more complexity to the provider framework.

Done = increase Lambda timeout to 60 seconds (not a great solution as it's horrible UX to have to wait 60s for options to load, but at least they will load at all if we have any slow endpoints like this one)
Done = use `errgroup.WithContext` to make the requests concurrently to the AWS SSO API. There are examples of this elsewhere in the codebase.
IO = you'll need to use a `sync.Mutex` when writing to `opts` to prevent different goroutines writing over one another. There are also examples of this elsewhere in the codebase

e.g.

```go
opts := []types.Option{}
var mu sync.Mutex

g, gctx := errgroup.WithContext(ctx)

g.Go(func() error {
// call the APIs to describe the permission set, etc
// use gctx as the context, rather than ctx, as gctx is cancelled automatically if there are errors.
// (read more about this in the errgroup package)
p.checkPermissionSetIsManagedByGranted(gctx, ...)
mu.Lock()
defer mu.Unlock()
opts = append(opts, types.Option{Label: *po.PermissionSet.Name, Value: arnCopy})
}
```

chrnorm:
Fixes #172

Also fixes a bug where `checkPermissionSetIsManagedByGranted` could loop forever, as we weren't passing `NextToken` into the API call.


chrnorm:
Implementing concurrency didn't speed things up enough unfortunately - the options request still takes about 7-15 seconds for me in local dev for AWS SSO Permission Sets.

To make this a lot faster I've implemented argument option caching in this PR. It does add a little bit of complexity but it makes the options *way* faster, they now take around 300ms and feel very snappy.

Here's a demo: https://www.loom.com/share/cca32c4658c649e5a4e6f15049c48752

meyerjrr (approved):

JoshuaWilkes:
Manish an I actually found the same outcome with trying to use goroutines for the option loading, it had no impact or worse performance.

There are at least 3 api calls per permission set which all get rate limited together.
We added a big slowdown with the tags lookup , that api just seems slower generally.

If we could pull off a provider that uses an assumed role into the management account, we could potentially drop the requirement for tags etc all together and just make all permission sets and accounts available.

PR #182, opened 2022-08-24
Provider setup UI

### LINKED ISSUE #194: Switch to semver for provider versioning
Done = we use https://pkg.go.dev/github.com/hashicorp/go-version for versioning providers

chrnorm:
Adds an interactive setup UI to run through setting up Access Providers.


accesshandler/pkg/providers/aws/sso/validate.go
```diff
@@ -84,3 +86,41 @@ func (p *Provider) ensureAccountExists(ctx context.Context, accountID string) er
 
 	return err
 }
+
+var _ providers.ConfigValidator = &Provider{}
+
+func (p *Provider) ValidateConfig() map[string]providers.ConfigValidationStep {
```
chrnorm, line 92: @JoshuaWilkes I've written a new testing interface which has a bit more functionality than `gconfig.TestConfig`.

The benefits of this approach over the existing one are:
- we can specify multiple config checks, each with an ID which can be provided to Common Fate for diagnostics if a check fails
- checks can emit logs to provide more insight into what is happening during a config test
- ability to show what the checks are for a particular object without actually running them

Are you happy with this approach? If so I am keen to shift this into `gconfig` to replace TestConfig, i.e.

```diff
// Tester interface is used to run tests on config
type Tester interface {
	// TestConfig is expected to be called on a loaded config after Init has been called.
-	TestConfig(ctx context.Context) error
+	ValidateConfig() map[string]ConfigValidationStep
}

File: cmd/gdeploy/commands/provider/add.go [hunk 2]
```diff
@@ -27,69 +30,92 @@ var addCommand = cli.Command{
...
+			if err != nil {
+				return err
+			}
 		}
-		var chosen string
-		err = survey.AskOne(&p, &chosen)
-		if err != nil {
-			return err
+
+		id := c.String("id")
```
JoshuaWilkes, line 61: Given the ID does show up in the UI for users, should we be keeping the significance here?
Or perhaps adding a friendly Name field?

cmd/gdeploy/main.go
```diff
@@ -67,7 +67,7 @@ func main() {
 			WithBeforeFuncs(&sso.SSOCommand, RequireDeploymentConfig(), VerifyGDeployCompatibility(), RequireAWSCredentials()),
 			WithBeforeFuncs(&backup.Command, RequireDeploymentConfig(), PreventDevUsage(), VerifyGDeployCompatibility(), RequireAWSCredentials()),
 			WithBeforeFuncs(&restore.Command, RequireDeploymentConfig(), PreventDevUsage(), VerifyGDeployCompatibility(), RequireAWSCredentials()),
-			WithBeforeFuncs(&provider.Command, RequireDeploymentConfig(), VerifyGDeployCompatibility(), RequireAWSCredentials()),
+			WithBeforeFuncs(&provider.Command, RequireDeploymentConfig(), VerifyGDeployCompatibility()),
```
JoshuaWilkes: We need to make sure that we still have consistent error messages when the user choosed the CLI based flow.

We could add the call to RequireAWSCredentials in the "add" command itself when the if block for prompt CLI is entered.

pkg/providersetup/step.go
```diff
@@ -0,0 +1,62 @@
...
+		SetupID:      setupID,
+		Index:        index,
+		Title:        s.Title,
+		Instructions: s.Instructions,
+	}
+	for _, field := range s.ConfigFields {
+		configField := types.ProviderConfigField{
+			Id:          field.Key(),
+			Name:        field.Key(),
+			Description: field.Usage(),
```
JoshuaWilkes: I was considering adding a description field in addition to Usage.

Usage is shown when rendering CLI prompts for a field, but we may want to add longer format descriptions too.
Also for the provider as a whole.

Do you think we have a naming issue here?

cmd/gdeploy/commands/provider/add.go
```diff
@@ -27,69 +30,92 @@ var addCommand = cli.Command{
...
+					err := deploy.CLIPrompt(v)
+					if err != nil {
+						return err
+					}
+				}
 			}
-			if dc.Deployment.Parameters.ProviderConfiguration == nil {
-				return nil
+
+			err = deploy.RunConfigTest(ctx, provider.Provider)
```
JoshuaWilkes, line 95: Leaving a note here that I think we will no longer be able to run these tests directly (and probably shouldn't)
When we use the assume role paradigm.

This is because the role used in the test should only have a trust relationship with the access handler execution roles.
So config tests should actually fail when the user is using a deployment role.

Because you have added an API to run these tests in the access handler, but this is not available to the CLI user either directly or via the approvals api due to auth.
We could potentially invoke the Access handler directly with a crafted request that mocks the API gateway.
Though it doesn't sound great. 

It's **critical** that we test any role relations as part of the test suite for a provider.

One solution is that tests only run via the web UI.
We can still using the cli prompt for dev/power user, but tests don't run in this workflow.

What do you think?

accesshandler/pkg/providers/aws/sso/validate.go
JoshuaWilkes: Looks good, Checkout my other comments as one is related to testing

cmd/gdeploy/main.go
chrnorm: Good call here, the reason I removed this from the `WithBeforeFuncs` stack was because it wasn't needed when using the UI based flow. I'll add it to the command itself when the CLI flow is used.

cmd/gdeploy/commands/provider/add.go
chrnorm: Good point. Agreed that tests will now need to run server-side moving forwards. 

One possibility here is to introduce a Deployment API which is authenticated via AWS IAM, and has very limited functionality. The API could provide the testing endpoint which would work similarly to the one used in the web UI flow. The CloudFormation stack could specify allowed roles which can access this API, and when running “gdeploy init” we’d set this parameter to be the user’s current role. 

What do you think of this?

Another alternative is that the providers could be represented as CloudFormation CRDs and then create/update operations run the tests. Not sure if this is the best approach though, as we’ll want to continuously check provider connections even if the provider itself isn’t being updated.

pkg/providersetup/step.go
chrnorm: What do you think of renaming “Usage” to “Description” @JoshuaWilkes in gconfig? I do think there is a naming issue and that “Description” would be fine (similar to how our Access Rule schemas have a Description field)

cmd/gdeploy/commands/provider/add.go
chrnorm: name field could be good - we could make it optional too. My thinking on the changes to “id” in this CLI command was to try and reduce the cognitive overhead for users by eliminating the question prompting for the provider ID. It should only be needed in advanced use cases where there are multiple copies of one provider installed.

I think showing the ID is useful (I know this isn’t related to your question but just sharing my POV here) as it makes it easy to see which access rules refer to which provider.

chrnorm:
@JoshuaWilkes Ill get this to handle secrets tomorrow and then this should be good to merge. Will also write very quick setup instructions for all our providers.

pkg/providersetup/step.go
JoshuaWilkes: Renaming to Description is ok

cmd/gdeploy/commands/provider/add.go
JoshuaWilkes: There a a few considerations here.

For non AWS configurables, including Identity providers for okta, azure etc that use the gconfig framework.
Testing can still happen in the CLI. (Though should we also be considering shifting these into the frontend setup workflow?)

For other non aws providers, the cli testing will work.
Its only the AWS providers where testing wont work without the assumed role.

I think we should add a condition to the tests so they only run in the cli if they are able to, and if they require the assumed role. A warning is returned instead.

If the user has gone the UI configuration route, they will have had the testing work for them.

So we can accept the poorer UX for now when using the CLI directly, which we expect our users wont need to do?

chrnorm:
Fixes #194

chrnorm:
To test this PR I recommend checking out the branch and running through the guided setup flow yourself - you can visit the "Providers" tab in the admin navbar to get there.

cmd/gdeploy/commands/provider/add.go
chrnorm: Long term I think doing all testing server side is preferable as it will generate logs for testing and it means that all requests to integrated services are coming from a team's cloud environment rather than a development laptop.

I think your approach flagged is good if you can make a ticket for implementing the condition and warning that would be fantastic. Currently there is no validation at all if you use the UI flow

chrnorm:
The interactive setup flow now writes sensitive values to AWS SSM. In order to get this to work I had to rework how our provider setup IDs were generated, and make them use the same ID format as the `granted-deployment.yml` config file. This is so that we can write secrets to paths like `/granted/providers/okta/apiToken` - we need to know the ID of the provider we're trying to write the secret to.

This has also simplified the setup flow as our API can check whether a provider setup with a particular ID also exists in the registered providers, and automatically remove the setup from the database once a deployment is complete.

JoshuaWilkes (approved):

PR #184, opened 2022-08-26
added grant id to instructioner interface and removed from args

### LINKED ISSUE #183: EKS provider has a 'grantId' field
This causes the provider to break - as noted by @meyerjrr

meyerjrr:
Fixes #183


jordiup (approved):
LGTM 👌

PR #185, opened 2022-08-26
One access rule for multiple targets

JoshuaWilkes:
Shift the caching into a service for reusability

Add selectable with fields, meaning that multiple targets can be selected when creating an access rule
Then the user can selected which one they want


pkg/rule/accessrule.go
```diff
@@ -136,9 +136,68 @@ func (a *Approval) IsRequired() bool {
 // I expect this will be different to what gets returned in the api response
 type Target struct {
 	// References the provider's unique ID
-	ProviderID   string            `json:"providerId"  dynamodbav:"providerId"`
-	ProviderType string            `json:"providerType"  dynamodbav:"providerType"`
-	With         map[string]string `json:"with"  dynamodbav:"with"`
+	ProviderID     string              `json:"providerId"  dynamodbav:"providerId"`
+	ProviderType   string              `json:"providerType"  dynamodbav:"providerType"`
+	With           map[string]string   `json:"with"  dynamodbav:"with"`
```
chrnorm, line 141: Rather than having both `With` and `WithSelectable`, could we not simply have `With` as a `map[string][]string`? It looks like this could cause us issues in future if we need to remember to check `With` as well as `WithSelectable` (and we might default to using `With` and forget to also check `WithSelectable`).

My thinking here is that a rule with a singular target e.g. `{"account": "123}` could be represented as an array of length 1, e.g. `{"account": ["123"]}`.

chrnorm (approved):
Looks good - I'm happy for this to go in. I'm thinking that adding `WithSelectable` introduces some complexity in the codebase, so let me know what you think of my comment @JoshuaWilkes (we can make this a separate task ticket if we want to remove `WithSelectable`)

pkg/rule/accessrule.go
chrnorm: I note that this might require a DynamoDB migration to run - but that might be a good opportunity to figure out how to execute a data migration script as part of a version upgrade.

JoshuaWilkes: I discussed it in my exploration doc, but my motives were compatibility, not wanting to have to work with migrations.
And I think it actually makes it a simpler distinction to be honest.

Though it probably looks a bit spaghetti code

PR #189, opened 2022-08-30
Providerkit POC

JoshuaWilkes:
proof of concept PR to explore how we can share some boiler plate AWS code for AWS providers into a providerkit library

This exampl shows how we can make the AWS SSO objects reusable


accesshandler/pkg/providerkit/aws/sso/sso.go
```diff
@@ -0,0 +1,58 @@
+package providerkitawsssov1
```
chrnorm: Thoughts on `pkit` as the package name over `providerkit`?

PR #190, opened 2022-08-30
ECS Flask provider V1

meyerjrr:


accesshandler/pkg/providers/flask/access.go
```diff
@@ -0,0 +1,361 @@
+package flask
+
```
meyerjrr, line 2: Cleanup to be done in this file I think. A lot of very similar code used here and in the eks provider

accesshandler/pkg/providers/flask/flask.go
```diff
@@ -0,0 +1,168 @@
...
+	"github.com/aws/aws-sdk-go-v2/service/organizations"
+	"github.com/aws/aws-sdk-go-v2/service/ssoadmin"
+	"github.com/aws/aws-sdk-go-v2/service/sts"
+	"github.com/common-fate/granted-approvals/accesshandler/pkg/types"
+	"github.com/common-fate/granted-approvals/pkg/gconfig"
+	"github.com/invopop/jsonschema"
+	"go.uber.org/zap"
+)
+
+type Provider struct {
```
meyerjrr, line 20: maybe we could embed a type in here to encapsulate all the clients that are needed in these sso providers.

accesshandler/pkg/providers/flask/flask.go
```diff
@@ -0,0 +1,168 @@
...
+	ssoRoleARN       gconfig.StringValue
+	ecsAccessRoleARN gconfig.StringValue
+
+	options map[string][]types.Option
+}
+
+func (p *Provider) Config() gconfig.Config {
+	return gconfig.Config{
+
+		gconfig.StringField("type", &p.providerType, "The type of the provider to display in the UI"),
```
meyerjrr: Do we need this, was in @chrnorm initial prototype

chrnorm: no we don't - this was only used for demo providers

chrnorm (changes requested):
@meyerjrr I just realised that this work has all been based on top of my prototype branch. My prototype branch used some modifications to `gconfig` to allow me to build demo providers more easily, which included changing a lot of types from `map[string]string` to `map[string]interface{}`. I don't think these changes are actually needed in order to implement Access Tokens for providers.

Could you please have a go at pulling the actual changes required to implement the Flask provider out of this branch and putting them on a new branch based on the latest `main` version?

We basically just need the Flask provider itself and the Access Token stuff.

PR #197, opened 2022-08-31
Add remaining CRUD subcmd for `gdeploy identity users` & `gdeploy identity groups` 

Eddie023:
### Description 

1. Add `delete` cmd to delete users from cognito identity provider.
2. Breakdown users file for specific cmds like `create` `delete`. 
3. Add `create` & `delete` subcmd to add or  delete user groups from cognito. 
4. Breakdown groups related files based on cmd name.


cmd/gdeploy/commands/identity/users/delete.go
```diff
@@ -0,0 +1,74 @@
...
+			return err
+		}
+
+		cog := cognitoidentityprovider.NewFromConfig(cfg)
+
+		_, err = cog.AdminRemoveUserFromGroup(ctx, &cognitoidentityprovider.AdminRemoveUserFromGroupInput{
+			GroupName:  &adminGroup,
+			UserPoolId: &o.UserPoolID,
+			Username:   &username,
+		})
```
Eddie023: Unsure if we need to delete users from `AdministratorGroupID` in `granted-deployment.yml`. Also, confused that this is not failing for users that were not added as admin i.e `gdeploy identity users -u abc@gmail.com`.

JoshuaWilkes: You should be able to add or remove any user from any group.
The command likely doesn't fail, but the response from cognito may be that no action was taken

Eddie023: In any case, this case should be added right? So that `- admin` users when deleted are removed from the `grantedAdministrator` group.

JoshuaWilkes: My assumption is that when a user is deleted in cognito, they are automatically remove from all groups.
Is that the issue you are talking about here?

Eddie023: Kind of. If that's the case then this condition is not necessary.

JoshuaWilkes (approved):

PR #206, opened 2022-09-01
ECS Exec provider

meyerjrr:
Adds the ECS Exec Access Provider. This provider provisions AWS SSO Permission Sets that allow users to shell in to tasks in an ECS cluster.


accesshandler/pkg/providers/aws/flask/access.go
```diff
@@ -0,0 +1,582 @@
...
+	})
+	if err != nil {
+		return err
+	}
+	// if the assignment deletion was not successful, return the error and reason
+	if status.AccountAssignmentDeletionStatus.FailureReason != nil {
+		return fmt.Errorf("failed deleting account assignment: %s", *status.AccountAssignmentDeletionStatus.FailureReason)
+	}
+
+	log.Info("Deleting  permission set", aws.String(p.instanceARN.Get()))
```
chrnorm: @meyerjrr can you replace your logging calls with `log.Infow`? This will make it easier to filter the logs.

```suggestion
	log.Infow("Deleting  permission set", "permissionset.arn" aws.String(p.instanceARN.Get()))
```

Your current log message will be printed like this:
```json
{ "msg": "Deleting permission set arn:::aws:permissionset/12345" }
```

If you use `log.Infow` as per the suggestion, your log message will be printed like this:
```json
{ "msg": "Deleting permission set", "permissionset.arn": "arn:::aws:permissionset/12345" }
```

Using structured fields makes it a lot easier to query log messages using CloudWatch or external tools like DataDog, rather than having to parse the `msg` field.

Essentially the guide here is to always use `log.Infow` when you have values that you'd like to include in log messages.

Note that you have to include a name for the field too, such as `permissionset.arn`.

There are a few instances to change this - could you run through the PR and update all of your log entries?

Here's another example from the [zap docs](https://pkg.go.dev/go.uber.org/zap#hdr-Choosing_a_Logger):
```
log.Infow("failed to fetch URL",
  "url", "http://example.com",
  "attempt", 3,
  "backoff", time.Second,
)
```

accesshandler/pkg/providers/aws/flask/access.go
```diff
@@ -0,0 +1,582 @@
...
+	// find the user ID from the provided email address.
+	user, err := p.getUser(ctx, subject)
+	if err != nil {
+		return nil, err
+	}
+	// create permission set with policy
+	permSet, err := p.ssoClient.CreatePermissionSet(ctx, &ssoadmin.CreatePermissionSetInput{
+		InstanceArn: aws.String(p.instanceARN.Get()),
+		Name:        aws.String(permissionSetName),
+		Description: aws.String("This permission set was automatically generated by Granted Approvals"),
```
chrnorm: We can make this more descriptive

```suggestion
		Description: aws.String("Granted Approvals ECS Flask access"),
```

accesshandler/pkg/providers/aws/flask/flask.go
```diff
@@ -0,0 +1,113 @@
...
+	ecsClient        *ecs.Client
+	ssoClient        *ssoadmin.Client
+	iamClient        *iam.Client
+	ssmClient        *ssm.Client
+	cloudtrailClient *cloudtrail.Client
+	idStoreClient    *identitystore.Client
+	orgClient        *organizations.Client
+	awsAccountID     string
+
+	// configured by gconfig
```
chrnorm: Add an extra space here to avoid Go associating this comment with the specific value below it (`ecsClusterARN`)
```suggestion
// the below fields are configured by gconfig

```

accesshandler/pkg/providers/aws/flask/flask.go
```diff
@@ -0,0 +1,113 @@
...
+	// a role which can be assumed and has required sso permissions
+	ssoRoleARN       gconfig.StringValue
+	ecsAccessRoleARN gconfig.StringValue
+
+	options map[string][]types.Option
+}
+
+func (p *Provider) Config() gconfig.Config {
+	return gconfig.Config{
+
```
chrnorm:
```suggestion
```

accesshandler/pkg/providers/aws/flask/flask.go
```diff
@@ -0,0 +1,113 @@
...
+// // Init the provider.
+func (p *Provider) Init(ctx context.Context) error {
+
+	//manually set the options for now
+	optionsJson := []types.Option{}
+
+	optionsJson = append(optionsJson, types.Option{Label: "ECS Demo", Value: "ecs-demo"})
+
+	p.options = make(map[string][]types.Option)
+	p.options["server"] = optionsJson
```
chrnorm: we should be able to delete this now

accesshandler/pkg/providers/aws/flask/flask.go
```diff
@@ -0,0 +1,113 @@
...
+	ssoCfg, err := config.LoadDefaultConfig(ctx, config.WithRegion(p.ssoRegion.Get()), config.WithCredentialsProvider(cfaws.NewAssumeRoleCredentialsCache(ctx, p.ssoRoleARN.Get(), cfaws.WithRoleSessionName("accesshandler-flask"))))
+	if err != nil {
+		return err
+	}
+	ecsCfg, err := config.LoadDefaultConfig(ctx, config.WithRegion(p.ecsRegion.Get()), config.WithCredentialsProvider(cfaws.NewAssumeRoleCredentialsCache(ctx, p.ecsAccessRoleARN.Get(), cfaws.WithRoleSessionName("accesshandler-flask"))))
+	if err != nil {
+		return err
+	}
+
+	// TODO: verify here if the ecs task has exec is enabled on the ecs task
```
chrnorm: flagging this as a review comment - this check should be implemented for sure

accesshandler/pkg/providers/aws/flask/setup/2.md
```diff
@@ -0,0 +1,89 @@
+---
+title: Create an IAM role
+configFields:
+  - ssoRoleARN
+  - ssoRegion
+  - clusterAccessRoleArn
+  - ecsRegion
+---
+
+# Create A SSO Management Role + Policy
```
chrnorm: Both of these policies need to be turned into CloudFormation, and need to include a trust relationship to the `AccessHandlerExecutionRoleARN`. You can look at the AWS SSO v2 provider for an example on how to do this.

accesshandler/pkg/providers/aws/flask/setup/3.md
```diff
@@ -0,0 +1,22 @@
...
+
+You can now test the connection!
+
+Afterwards you will be able to:
+- Create a Flask access rule
+- Create an access request for the Flask deployment
+- Assume into the ECS cluster
+
+To finalise the setup you will need to install [granted-flask](https://pypi.org/project/granted-flask/)
+-  More on this here [docs.commonfate.io/granted-approvals/providers/flask](https://docs.commonfate.io/granted-approvals/providers/flask)
```
chrnorm: Move the contents of docs.commonfate.io/granted-approvals/providers/flask into this step rather than linking out to a separate website

accesshandler/pkg/providers/aws/flask/setup/3.md
```diff
@@ -0,0 +1,22 @@
...
+
+Once you have updated your Flask deployment with the correct webhook URL you can test your connection with the `test` command
+Run the following to test your configuration is correctly setup:
+```
+aws ecs execute-command --cluster {CLUSTER_ARN} --task {TASK_ID} --container DefaultContainer --interactive --command 'flask test'
```
chrnorm: I like the test command but we shouldn't make this a flask command unless we're overriding functionality.

Can we replace this with `granted test` instead?

accesshandler/pkg/providers/aws/flask/testdata/argschema.json
```diff
@@ -0,0 +1,16 @@
+{
```
chrnorm: This argschema looks like it needs to be updated - it should be used in tests to check that the schema of the provider matches what we expect.

accesshandler/pkg/providers/aws/flask/validate.go
```diff
@@ -0,0 +1,63 @@
...
+				res, err := p.idStoreClient.ListUsers(ctx, &identitystore.ListUsersInput{
+					IdentityStoreId: aws.String(p.identityStoreID.Get()),
+				})
+				if err != nil {
+					return diagnostics.Error(err)
+				}
+				return diagnostics.Info("AWS SSO returned %d users (more may exist, pagination has been ignored)", len(res.Users))
+			},
+		},
+		"assume-role": {
```
chrnorm: This needs to be replaced with a test that actually tries to assume a role - see the AWS SSO v2 provider as this has now been fixed there.

accesshandler/pkg/providers/aws/flask/validate.go
```diff
@@ -0,0 +1,63 @@
...
+		AccountId: &accountID,
+	})
+	var anf *orgtypes.AccountNotFoundException
+	if errors.As(err, &anf) {
+		return &AccountNotFoundError{AccountID: accountID}
+	}
+
+	return err
+}
+func (p *Provider) ValidateConfig() map[string]providers.ConfigValidationStep {
```
chrnorm: Can you please add some extra config validations?

- try to assume both the cluster access role as well as the SSO access role (should be two separate checks)
- try to describe tasks in the cluster

accesshandler/pkg/providers/providers.go
```diff
@@ -20,6 +20,12 @@ type Accessor interface {
 	Revoke(ctx context.Context, subject string, args []byte, grantID string) error
 }
 
+// If a provider implements RequiresAccessToken
+// It indicates that an access token should be generated for use by this provider.
+type RequiresAccessToken interface {
+	RequiresAccessToken()
```
chrnorm: an empty function like this looks a bit weird - let's make this return a boolean. For access providers which use this we can return `true`.
```suggestion
	RequiresAccessToken() bool
```

accesshandler/pkg/providers/aws/flask/setup/1.md
```diff
@@ -0,0 +1,42 @@
...
+```bash
+> aws sso-admin list-instances
+{
+    "Instances": [
+        {
+            "InstanceArn": "arn:aws:sso:::instance/ssoins-825968feece9a0b6",
+            "IdentityStoreId": "d-976708da7d"
+        }
+    ]
+}
```
chrnorm: These point to our actual sandbox org - better to replace them with values which are more obviously placeholders.

```suggestion
❯ aws sso-admin list-instances
{
    "Instances": [
        {
            "InstanceArn": "arn:aws:sso:::instance/ssoins-1234567890",
            "IdentityStoreId": "d-1234567890"
        }
    ]
}
```

cmd/lambda/webhook/handler.go
```diff
@@ -2,43 +2,200 @@ package main
...
+	r.Post("/webhook/v1/test-setup", func(w http.ResponseWriter, r *http.Request) {
+
+		//successfull connection to webhook url return OK
+		w.WriteHeader(http.StatusOK)
+
+	})
+
+	r.Post("/webhook/v1/events-recorder", func(w http.ResponseWriter, r *http.Request) {
+		ctx := r.Context()
+		token := r.Header.Get("X-Granted-Request")
```
chrnorm: let's change this - the original `X-Granted-Request` contained the request ID which we've now shifted away from to Access Tokens.

```suggestion
		token := r.Header.Get("X-CommonFate-Access-Token")
```

chrnorm: Note that this will require updating the Python library

cmd/lambda/webhook/handler.go
```diff
@@ -2,43 +2,200 @@ package main
...
+		err = ValidateToken(ctx, t, q)
+		if err != nil {
+			apio.Error(ctx, w, err)
+			return
+		}
+		w.WriteHeader(http.StatusOK)
+
+	})
+
+	r.Post("/webhook/v1/test-setup", func(w http.ResponseWriter, r *http.Request) {
```
chrnorm: If we're just returning a 200, this is basically a healthcheck endpoint.

```suggestion
	r.Post("/webhook/v1/health", func(w http.ResponseWriter, r *http.Request) {
```

cmd/lambda/webhook/handler.go
```diff
@@ -2,43 +2,200 @@ package main
...
+
+		//successfull connection to webhook url return OK
```
chrnorm:
```suggestion
		//successful connection to webhook url return OK
```

cmd/lambda/webhook/handler.go
```diff
@@ -50,3 +207,43 @@ func (h *Lambda) Handler(ctx context.Context, req events.APIGatewayProxyRequest)
...
+		return nil, err
+	}
+
+	return s, nil
+}
+
+func (s *Server) Run(ctx context.Context) error {
+
+	serv := &http.Server{
+		Addr:    "localhost:3030",
```
chrnorm: As per my comment above in the Lambda file this should be shifted out - we shouldn't see any `localhost` references here as this file is for AWS Lambda

pkg/access/access_token.go
```diff
@@ -0,0 +1,35 @@
...
+import (
+	"time"
+
+	"github.com/common-fate/ddb"
+	"github.com/common-fate/granted-approvals/pkg/storage/keys"
+	"github.com/common-fate/granted-approvals/pkg/types"
+)
+
+type AccessToken struct {
+	RequestId string `json:"request_id"` // maybe?
```
chrnorm: Feedback on this section:
- everything we write to the DB needs to have a `dynamodbav` tag on all struct fields
- for JSON and dynamodbav tags we always use camelCase
- for Go variables we should never use `somethingId`, it should always be `somethingID` (the only case where this is acceptable in the codebase is in the auto-generated OpenAPI structs)

```suggestion
	RequestID string `json:"requestId" dynamodbav:"requestId"`
```

pkg/access/access_token.go
```diff
@@ -0,0 +1,35 @@
...
+	Token     string `json:"token"`
```
chrnorm:
```suggestion
	Token     string `json:"token" dynamodbav:"token"`
```

pkg/storage/get_access_token.go
```diff
@@ -0,0 +1,37 @@
...
+	"github.com/aws/aws-sdk-go-v2/feature/dynamodb/attributevalue"
+	"github.com/aws/aws-sdk-go-v2/service/dynamodb"
+	"github.com/aws/aws-sdk-go-v2/service/dynamodb/types"
+	"github.com/common-fate/ddb"
+	"github.com/common-fate/granted-approvals/pkg/access"
+	"github.com/common-fate/granted-approvals/pkg/storage/keys"
+)
+
+type GetAccessToken struct {
+	ID     string
```
chrnorm: can be more specific here

```suggestion
	RequestID     string
```

pkg/recorder/event.go
```diff
@@ -0,0 +1,25 @@
+package recorder
+
+import (
+	"time"
+
+	"github.com/common-fate/ddb"
+	"github.com/common-fate/granted-approvals/pkg/storage/keys"
+)
+
+type Event struct {
```
chrnorm: We should be able to delete this right? I think I added it in my initial prototype before I realised we could reuse the `RequestEvent` struct for it.

pkg/storage/keys/recorderevent.go
```diff
@@ -0,0 +1,17 @@
+package keys
+
+const RecorderEventKey = "RECORDER_EVENT#"
```
chrnorm: should be able to delete this

pkg/types/resource_id.go
```diff
@@ -39,3 +39,7 @@ func NewHistoryID() string {
 func NewProviderSetupID() string {
 	return newResourceID("pse")
 }
+
+func NewRecorderEventID() string {
+	return newResourceID("evt")
+}
```
chrnorm: should be able to delete this as it's unused
```suggestion
```

web/src/components/Request.tsx
```diff
@@ -265,34 +273,107 @@ export const RequestAccessInstructions: React.FC = () => {
...
+
   if (!data || !data.instructions) {
     return null;
   }
 
+  if (reqData?.grant?.status === "PENDING") {
+    return (
+      <Stack>
+        <Box textStyle="Body/Medium" mb={2}>
+          Granted is provisioning your access, it'll be done soon!
```
chrnorm: we can remove "your" here and make this a bit more concise - if reviewers are looking at the page it's not their access which is being provisioned.

```suggestion
          Provisioning access
```

web/src/components/Request.tsx
```diff
@@ -265,34 +273,107 @@ export const RequestAccessInstructions: React.FC = () => {
...
+
   return (
     <Stack>
       <Box textStyle="Body/Medium" mb={2}>
         Access Instructions
       </Box>
       return (
       <ReactMarkdown
+        // remarkPlugins={[remarkGfm]}
+        skipHtml={false}
```
chrnorm: Is `skipHtml` required? I'm concerned that if we start allowing arbitrary HTML in the setup instructions we'll expose the app to XSS vulnerabilities.

```suggestion
```

accesshandler/pkg/api/grants.go
```diff
@@ -19,31 +21,40 @@ func (a *API) PostGrants(w http.ResponseWriter, r *http.Request) {
...
+			apio.Error(ctx, w, err)
+			return
 		}
+		return
+	}
 
-		return a.runtime.CreateGrant(ctx, *g)
-	}()
-
+	grant, additionalProperties, err := a.runtime.CreateGrant(ctx, *g)
```
chrnorm: I understand the approach you're taking here but I'm not convinced that adding `additionalProperties` to our Access Provider framework is a good idea.

It seems that `additionalProperties` is used as a generic output mechanism where the Access Provider has generated some data which we wish to store. Currently our only use case for `additionalProperties` is allowing the flask provider to output an Access Token.

However, Access Tokens are not just any old data - they have special meaning and need to be stored in our database so that our webhook handler can look it up. Using `additionalProperties` for this means that we're introducing a generic framework feature which can be confusing (as an Access Provider developer, the question I'd ask is - when should I be using `additionalProperties`), which at the same time has specific meaning.

This approach has led to the Access Token implementation being spread across 3 parts of the codebase:
- the Approvals API (handling the Access Token in the `additionalProperties` response)
- the Access Handler API (generating the Access Token)
- the provider itself (implementing `RequiresAccessToken`)

On the [YAGNI principle](https://martinfowler.com/bliki/Yagni.html) I don't think we should take this approach. My main opposition is that we're introducing a generic "additionalProperties" field without defined use cases on how this will be used across many Access Providers.

Here's my proposed approach which I think makes our framework simpler:

- when creating a request, the Approvals API checks whether the provider implements `RequiresAccessToken`. If it does, the Approvals API generates the Access Token and saves it in the database.

This requires the Approvals API to know about which providers are installed - but it already does! The API now has access to the provider metadata and can look up a provider without needing to fully initialise it. You can have a look at the `providersetup` endpoints I wrote which utilise this.

Separately, I do think the idea of having storage associated with a Grant can be useful, and I think we'll return to this to design this out in future once we have some concrete use cases in mind. I don't think Access Tokens are part of this generic storage though.

chrnorm (dismissed):
Looking good @meyerjrr! My biggest piece of feedback is around the introduction of `additionalProperties` here, see the review comment I've left for my proposed alternative. Other than that the rest of my feedback is on syntax/language stuff.

accesshandler/pkg/providerregistry/registry.go
```diff
@@ -43,6 +44,12 @@ func (pr ProviderRegistry) All() map[string]RegisteredProvider {
 func Registry() ProviderRegistry {
 	return ProviderRegistry{
 		Providers: map[string]map[string]RegisteredProvider{
+			"commonfate/flask": {
+				"v1": {Provider: &flask.Provider{},
```
chrnorm: Let's mark this as an alpha version too

```suggestion
				"v1-alpha1": {Provider: &flask.Provider{},
```

accesshandler/pkg/providers/aws/ecs-shell-sso/provider.go
```diff
@@ -0,0 +1,119 @@
...
+	if err != nil {
+		return err
+	}
+	if res.Account == nil {
+		return errors.New("aws accountID was nil in sts get caller id response")
+	}
+	p.awsAccountID = *res.Account
+
+	//check to see if cluster is running and has exec enabled
+	clusters, err := p.ecsClient.DescribeClusters(ctx, &ecs.DescribeClustersInput{Clusters: []string{p.ecsClusterARN.Get()}})
```
JoshuaWilkes, line 96: @chrnorm There is a check here that can fail if the state of the cluster changes.
The problem I see is that because this happens during Init, if it fails it will prevent the Access handler from ever starting, causing a complete outage of Approvals.

While its obvious that grants will fail if this is not enabled. I think this shows that we need to update our provider initialisation to make individual providers as failed rather than failing to start the access handler service at all.

As a start, I will shift this check into a validation that runs during provider setup, and we can ticket the work separately to improve the stability of the access handler

JoshuaWilkes: For this release I might just leave this as is to get it out

JoshuaWilkes (approved):

PR #209, opened 2022-09-01
Copy diagnostics button should copy logs to clipboard 

### LINKED ISSUE #193: Copy diagnostics button in setup flow doesn't work
Image

Done = pressing the 'copy diagnostics' button should copy the test results.
DE = the test results should be copied as markdown, so that they render nicely when pasted into Slack for debugging.

Eddie023:
### Loom 
https://www.loom.com/share/2ae587908dd844c58d341b3d68eea874


web/src/pages/admin/providers/setup/[id]/index.tsx
```diff
@@ -114,6 +118,10 @@ const Page = () => {
     if (data != null) {
       await mutate({ ...data, status: "VALIDATING" }, { revalidate: false });
       const res = await validateProvidersetup(id);
+
+      if (res?.configValidation.length > 0) {
```
Comment (JoshuaWilkes) [inline web/src/pages/admin/providers/setup/[id]/index.tsx:123]: I think this might be better as a useEffect
It should reset the state if the length is 0

JoshuaWilkes (changes requested):

web/src/pages/admin/providers/setup/[id]/index.tsx
Comment (JoshuaWilkes) [reply web/src/pages/admin/providers/setup/[id]/index.tsx:123]: Disregard this! my mistake

JoshuaWilkes (approved):

PR #210, opened 2022-09-01
Setup provider api errors

JoshuaWilkes:
This PR fixes a bug where config values were being lost. the issue was caused by a few FE and BE issues in combination.

This also addresses some apoi.Errror responses where the retrun statement was missing, leading to panics

Also addresses the handling of provider.Init failing during validation.
When the provider init fails, for example due to a role arn being misconfigured, the failure will now be reported as the result for all tests, before it would not report back to the fe client at all!

### DIFF
--- a/web/src/pages/admin/providers/setup/[id]/index.tsx
+++ b/web/src/pages/admin/providers/setup/[id]/index.tsx
@@ -75,7 +75,6 @@ const Page = () => {
   const [accordionIndex, setAccordionIndex] = useState([0]);
   const { data, mutate } = useGetProvidersetup(id);
   const { data: instructions } = useGetProvidersetupInstructions(id);
-  console.log({ data });
   // used to look up extra details like the name
   const registeredProvider = registeredProviders.find(
     (rp) => rp.type === data?.type
@@ -436,7 +435,6 @@ const StepDisplay: React.FC<StepDisplayProps> = ({
   const { control, handleSubmit } = useForm<Record<string, string>>({
     defaultValues: configValues,
   });
-  console.log({ step: step, configValues });
   const onSubmit = async (data: Record<string, string>) => {
     const filteredData: Record<string, string> = {};
     step.configFields.forEach((f) => {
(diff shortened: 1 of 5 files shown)


JoshuaWilkes:
fixes #207

web/src/pages/admin/providers/setup/[id]/index.tsx
```diff
@@ -436,13 +436,19 @@ const StepDisplay: React.FC<StepDisplayProps> = ({
...
     defaultValues: configValues,
   });
-
+  console.log({ step: step, configValues });
   const onSubmit = async (data: Record<string, string>) => {
+    const filteredData: Record<string, string> = {};
+    step.configFields.forEach((f) => {
+      filteredData[f.id] = data[f.id];
+    });
+    console.log({ datas: data, filteredData });
```
Comment (Eddie023) [inline web/src/pages/admin/providers/setup/[id]/index.tsx]:
```suggestion
    
```

web/src/pages/admin/providers/setup/[id]/index.tsx
```diff
@@ -75,7 +75,7 @@ const Page = () => {
   const [accordionIndex, setAccordionIndex] = useState([0]);
   const { data, mutate } = useGetProvidersetup(id);
   const { data: instructions } = useGetProvidersetupInstructions(id);
-
+  console.log({ data });
```
Comment (Eddie023) [inline web/src/pages/admin/providers/setup/[id]/index.tsx:78]: .

Comment (JoshuaWilkes) [inline web/src/pages/admin/providers/setup/[id]/index.tsx]:
```suggestion
```

web/src/pages/admin/providers/setup/[id]/index.tsx
```diff
@@ -436,13 +436,19 @@ const StepDisplay: React.FC<StepDisplayProps> = ({
   const { control, handleSubmit } = useForm<Record<string, string>>({
     defaultValues: configValues,
   });
-
+  console.log({ step: step, configValues });
```
Comment (JoshuaWilkes) [inline web/src/pages/admin/providers/setup/[id]/index.tsx]:
```suggestion
```

PR #211, opened 2022-09-01
Port setup docs

JoshuaWilkes:
This includes a bug fix for provider setups when secrets are included.

The bug was in gconfig Load() where it always treated values as having changed!


chrnorm (approved):
Looks good - I was having to work around that bug in `gconfig` in my initial setup docs implementation (I thought it was just the expected behaviour) so love the improvement here.

PR #212, opened 2022-09-01
add a gdeploy flag to ignore the version mismatch

chrnorm:
When creating UAT deployments it's a lot more convenient to build `gdeploy` from source, especially when we're confident that our YAML schema hasn't changed between versions. 

When doing prerelease testing I usually test the upgradability from the previous release to the current one, so I'll create an initial YAML with a tagged release (e.g. `v0.3.2`) and then update the deployment to point to the latest prerelease from `main` (e.g. `https://granted-test-releases-us-west-2.s3.amazonaws.com/dev/a9dcefa583baed5f6c3c7d94948269302383adf4/Granted.template.json`). However the VerifyGDeployCompatibility no longer allows me to use tagged releases when building `gdeploy` from source:

```
❯ gdeploy create
? Incompatible gdeploy version. Expected v0.3.2 got dev .
 Would you like to update your 'granted-deployment.yml' to make release version equal to  dev No
[✘] Please update gdeploy version to match your release version in 'granted-deployment.yml'.
```

This PR allows a `--ignore-version-mismatch` flag to be provided which skips the version check but displays warnings around its usage.

### DIFF
--- a/cmd/gdeploy/main.go
+++ b/cmd/gdeploy/main.go
@@ -32,6 +32,7 @@ func main() {
 		Flags: []cli.Flag{
 			&cli.PathFlag{Name: "file", Aliases: []string{"f"}, Value: "granted-deployment.yml", Usage: "The deployment configuration yml file path"},
 			&cli.BoolFlag{Name: "ignore-git-dirty", Usage: "Ignore checking if this is a clean repository during create and update commands"},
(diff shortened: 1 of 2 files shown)


JoshuaWilkes:
I didn't think about this UAT flow, the UAT we had supported was using http release URLS rather than version numbers

JoshuaWilkes (approved):

PR #214, opened 2022-09-01
add aliases for top-level gdeploy commands

chrnorm:
helps retain backwards compatibility and avoid a breaking change with the 'providers' command


JoshuaWilkes (approved):
We could consider shorthand aliases too 

gdeploy no slack ...
gdeploy pr add
gdeploy id users ...

PR #229, opened 2022-09-02
handle cases where permissionset access is restricted

chrnorm:
don't include Permission Sets that we can't describe in the aws-sso options


chrnorm:
Partial fix for #230 - needs some further testing.

PR #234, opened 2022-09-05
use a random string for the cognito domain prefix to avoid collisions

JoshuaWilkes:
This PR replaces the prompt for a cognito domain prefix with a random string.
This will reduce cognitive load for deployments and reduce the chance of any naming collision

It is backwards compatible, existing deployments are unaffected. the scope is only to the Init command.
You can optionally still specify a domain prefix via cli flag or via the config file directly.

### DIFF
--- a/pkg/deploy/config.go
+++ b/pkg/deploy/config.go
@@ -555,23 +555,11 @@ func SetupReleaseConfig(c *cli.Context) (*Config, error) {
 	// set up stack parameters
 	cognitoPrefix := c.String("cognito-domain-prefix")
 	if cognitoPrefix == "" {
-		var company string
-		p := &survey.Input{Message: "Your company name"}
-		err := survey.AskOne(p, &company)
-		if err != nil {
-			return nil, err
-		}
-		// turn the company name into lowercase and remove spaces, so that it will work
-		// as a Cognito prefix domain
-		company = strings.ReplaceAll(company, " ", "")
-		company = strings.ToLower(company)
-
-		cognitoPrefix = fmt.Sprintf("granted-login-%s", company)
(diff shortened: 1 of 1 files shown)


chrnorm (changes requested):
@JoshuaWilkes I like how this removes the need to ask a question during setup. For internal testing this is also nice as it avoids collisions. However the pain here is that Cognito does not support updating the hosted UI URL after an initial deployment. So the challenge I see with this approach is as follows:

- User is testing out Granted Approvals and creates an initial deployment
- User is happy with how Granted Approvals works and wants to productionise the deployment
- User enables SAML SSO
- User wants to change the Cognito URL to something recognisable to minimise phishing risk: with a randomly generated URL it's harder for end users to tell whether they are on the correct login URL. The threat scenario here is that a threat actor sets up their own Cognito user pool with a similar autogenerated URL, but linking to a fake SSO flow ([example](https://github.com/praetorian-inc/Okta_Watering_Hole))
- User needs to tear down and redeploy their whole Granted Approvals deployment to change the login URL 😠

The best solution to this is to allow users to specify a custom login URL for the Cognito UI (e.g. https://login.granted.mycompany.com). This requires us to add support for this in our CloudFormation, and I do think that the `granted-login-mycompany.auth.us-west-2.amazoncognito.com` is a good compromise for now.

What I propose is:
- for now, we make a change to the init logic to remove alphanumeric characters from the prefix (something like [this](https://www.golangprograms.com/replace-any-non-alphanumeric-character-sequences-with-a-dash-using-regex.html)). This solves the current issue where using special characters for their company name causes the initial deployment to fail.
- add an issue to support adding custom login domains after an initial stack has been deployed.
- when we have support for custom login domains, implement the functionality in this PR to remove the question for "Your company name" and use autogenerated login URLs.

PR #253, opened 2022-09-06
managed config for access providers

chrnorm:
## Overview

This PR introduces Managed Deployment Configuration for the Granted Approvals deployment.

In a self-hosted deployment, configuration of notification integrations like Slack and Access Providers is done in the `granted-deployment.yml` file. In a deployment hosted by Common Fate, we will manage the infrastructure itself and this YAML file is not required. However, users still need a way to configure their deployment with Access Providers and notifications! This is where Managed Deployment Configuration comes in.

Managed Deployment Configuration allows the Granted Approvals API to use designated DynamoDB entries as a source of truth for notifications and provider configuration, rather than relying on CloudFormation parameters.

This PR adds an interface to read deployment config:
```go
type DeployConfigReader interface {
	ReadProviders(ctx context.Context) (ProviderMap, error)
	ReadNotifications(ctx context.Context) (FeatureMap, error)
}
```

This interface is satisfied by both the `EnvAppConfig` struct (which reads config from the existing environment variables we are already using in self-hosted deployments) and the `DDBManagedDeploymentConfig` struct, which reads config from `MANAGED_DEPLOYMENT_CONFIG#PROVIDERS` and `MANAGED_DEPLOYMENT_CONFIG#NOTIFICATIONS` in DynamoDB. The `DDBManagedDeploymentConfig` struct is also capable of writing deployment config back to the database.

## Considerations

This PR uses DynamoDB as a source for reading and writing Managed Deployment Config. In order to allow the Access Handler service to load the config when Managed Deployment Config is enabled, the Access Handler has been granted read access to the specific items in the DynamoDB table. I've used a least-privilege role here so it only has read access over these keys, and nothing else. This is a change over our previous deployment, in which the Access Handler did not have any access at all to DynamoDB.

The initial approach here that @JoshuaWilkes and I discussed was using AWS SSM. This is nice as it doesn't feel like we're coupling our services together. However, when testing this I found that AWS SSM had quite significant latency which caused operations to take an extra 500ms+. The reasons I'm proposing we use DynamoDB here are:
- we already have the infrastructure in place as we can use the same table
- it's significantly faster than AWS SSM
- long term I'm proposing that we shift the workflow component of our architecture to be closer to the Approvals REST API. Because of this, coupling to DynamoDB may not be an issue.
image

We also discussed implementing a build-time flag for this. This PR uses an environment variable with a CloudFormation parameter rather than a build-time flag. My justification here is that it allows us to use the same release versions with self-hosted and and Managed Deployment Config versions hosted by us, which avoids having separate release processes for self-hosted vs Managed Deployment Config.


pkg/deploy/cfg_loader_ssm.go
```diff
@@ -0,0 +1,96 @@
...
+
+var (
+	managedConfigProviderPath      = "/commonfatecloud/config/providers"
+	managedConfigNotificationsPath = "/commonfatecloud/config/notifications"
+)
+
+var _ DeployConfigWriter = &SSMAppConfig{}
+
+// SSMAppConfig reads config values from environment variables.
+type SSMAppConfig struct {
```
chrnorm: This is unused - will be deleted before this PR is made ready for review

accesshandler/cmd/gdk/commands/fixtures/create.go
```diff
@@ -1,87 +0,0 @@
-package fixtures
```
chrnorm, line 1: I'll add this back in - didn't realise it was used in our CI testing

deploy/infra/lib/constructs/database.ts
```diff
@@ -0,0 +1,90 @@
...
+    approvals.addGlobalSecondaryIndex(gsi2);
+    approvals.addGlobalSecondaryIndex(gsi3);
+    approvals.addGlobalSecondaryIndex(gsi4);
+
+    const cfnTable = approvals.node.defaultChild as dynamodb.CfnTable;
+
+    // this table was previously part of the 'app-backend' construct.
+    // to avoid recreating the table as part of a deployment, we force
+    // the logical ID to be the same as the previous one.
+    cfnTable.overrideLogicalId("APIDBTableA8FD77F9");
```
JoshuaWilkes, line 82: will this work for everyone?

JoshuaWilkes (commented):
I couldn't really tell, does the dynamo config store in the same format with secrets as SSM params?

JoshuaWilkes:
On the point of latency

If we store configuration in dynamo db that includes secrets which are stored in SSM
then we need to do a query to dynamo followed by a query to SSM for each secret.

If we store the raw config in SSM we only do one trip to SSM.

This is nuance for providers which require secrets and Slack which has a secret.
Okta and Azure would be affected too.

deploy/infra/lib/constructs/database.ts
chrnorm: Yep it will - the logical ID is the same for everyone who is using the stack. However this needs to be specifically included in our UAT test plan if this is merged just to double check.

chrnorm:
@JoshuaWilkes RE latency - I believe we're stuck with doing at least 2 sequential calls to AWS SSM if we were to store the Managed Deployment Config there. I've put together a diagram below showing this.

We can (and do) resolve all of the secrets concurrently (e.g. replace a reference to `awsssm://my-secret-value:1` in the config file with the secret loaded from SSM). However we need to first fetch the Managed Deployment Config to determine which providers exist, so we know what secrets to load.

image

In my testing running commands like `time aws ssm get-parameter --name /christest` it looks like each call to SSM takes 500ms.

So the difference here is around 470ms if DynamoDB is used for the Managed Deployment Config.

However as you note - we still do need to call SSM to resolve the secrets such as Okta API tokens, Slack tokens etc.

chrnorm:
> I couldn't really tell, does the dynamo config store in the same format with secrets as SSM params?

@JoshuaWilkes yep it does. It's the same format as the YAML file, but saved as a DynamoDB map item rather in than YAML.

chrnorm:
Closing for now - I merged the interface improvements in #261 but excluded the DynamoDB managed config stuff.

PR #255, opened 2022-09-07
added loading bar for when grant is not active

meyerjrr:
Added loading indicator when grant is provisioning

### DIFF
--- a/web/src/components/Request.tsx
+++ b/web/src/components/Request.tsx
@@ -317,7 +317,7 @@ export const RequestAccessInstructions: React.FC = () => {
                 {props.children}
               </Text>
             ),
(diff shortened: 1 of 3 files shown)


web/src/components/Request.tsx
```diff
@@ -268,30 +270,63 @@ export const RequestAccessInstructions: React.FC = () => {
...
+    {
+      swr: {
+        refreshInterval: refreshInterval,
+      },
+    }
+  );
+
+  useEffect(() => {
+    if (reqData?.grant?.status == "PENDING") {
+      setRefreshInterval(1000);
```
JoshuaWilkes:
```suggestion
      setRefreshInterval(5000);
```

PR #257, opened 2022-09-07
okta: handle cases where usernames don't match email

chrnorm:
In some Okta tenancies, users have a username in the format `firstname.lastname` which doesn't match their email address. Granted Approvals uses email address to lookup users. This PR uses the Okta [List Users With Search](https://developer.okta.com/docs/reference/api/users/#known-limitation) API to match users by their email address.


JoshuaWilkes (approved):
Looks good

PR #258, opened 2022-09-08
Add Remove & Update subcommand for providers 

Eddie023:
### Description
1. Add `gdeploy update providers` & `gdeploy remove providers` subcmds. 
2. Fix issue when invalid `with=` arg is provided in `add` cmd. 

### Note
1. When providing `--with apple=orange` no check is done. So, essentially if users want to add dumb key=value, it will override the previous values for that provider.

### DIFF
--- a/pkg/deploy/config.go
+++ b/pkg/deploy/config.go
@@ -93,7 +93,7 @@ func (f *ProviderMap) Add(id string, p Provider) error {
 // Update the Provider if it exist
 func (f *ProviderMap) Update(id string, p Provider) error {
 	if _, ok := (*f)[id]; !ok {
-		return fmt.Errorf("provider %s doesnot exists in the config", id)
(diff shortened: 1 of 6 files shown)


pkg/deploy/config.go
```diff
@@ -90,6 +90,16 @@ func (f *ProviderMap) Add(id string, p Provider) error {
 	return nil
 }
 
+// Update the Provider if it exist
+func (f *ProviderMap) Update(id string, p Provider) error {
+	if _, ok := (*f)[id]; !ok {
+		return fmt.Errorf("provider %s doesnot exists in the config", id)
```
chrnorm:
```suggestion
		return fmt.Errorf("provider %s not found in config", id)
```

chrnorm (approved):
Looks good @Eddie023!

PR #260, opened 2022-09-08
Add user management features to web UI

JoshuaWilkes:
Adds user and group management for cognito via the web UI

image
image
image
image
image


pkg/api/request.go
```diff
@@ -311,7 +311,7 @@ func (a *API) RevokeRequest(w http.ResponseWriter, r *http.Request, requestID st
 		apio.Error(ctx, w, err)
 		return
 	}
-	apio.JSON(ctx, w, res, http.StatusOK)
+	apio.JSON(ctx, w, nil, http.StatusOK)
```
chrnorm, line 314: Is this change intentional?

pkg/service/cognitosvc/cognito_test.go
```diff
@@ -0,0 +1,3 @@
+package cognitosvc
```
chrnorm: We can just remove this file altogether IMO - otherwise it gives the impression that tests do exist if you're looking through the file structure in this repo.

pkg/api/request.go
JoshuaWilkes: Yes, I was reviewing th eAPI responses to check they matched the openapi spec, this one was returning a raw request when the spec states no response body.

I found some other instances where .ToAPI() was missing and fixed those up too.

This api probably should return something, but I figure it should at least match the schema!

chrnorm: Good call!

openapi.yml
```diff
@@ -543,6 +543,42 @@ paths:
...
+  "/api/v1/admin/users/{userId}":
+    parameters:
+      - schema:
+          type: string
+        name: userId
+        in: path
+        required: true
+    post:
+      summary: Update User
+      operationId: post-api-v1-admin-users-userId
```
chrnorm: nitpick: we should give these operations shorter names, so that the generated React Hooks and Go endpoints are more concise.

```suggestion
      operationId: update-user
```

openapi.yml
```diff
@@ -810,6 +884,32 @@ paths:
...
+          description: OK
+        "401":
+          $ref: "#/components/responses/ErrorResponse"
+        "500":
+          $ref: "#/components/responses/ErrorResponse"
+      description: Run the identity sync operation on demand
+  /api/v1/admin/identity:
+    get:
+      summary: Get identity configuration
+      tags: []
```
chrnorm: nitpick - these should be tagged with the "Admin" tag so they get code-generated into the right frontend file

openapi.yml
```diff
@@ -1775,6 +1880,20 @@ components:
                   $ref: "#/components/schemas/ProviderSetup"
             required:
               - providerSetups
+    IdentityConfigurationResponse:
+      description: Example response
```
chrnorm:
```suggestion
      description: Returns information about the identity configuration of this deployment.
```

pkg/server/routes.go
```diff
@@ -18,7 +18,7 @@ func (c *Server) Handler() http.Handler {
 	r.Use(c.requestIDMiddleware)
 	r.Use(chiMiddleware.RealIP)
 	r.Use(chiMiddleware.Recoverer)
-	r.Use(chiMiddleware.Timeout(30 * time.Second))
+	r.Use(chiMiddleware.Timeout(3000 * time.Second))
```
chrnorm: Could consider omitting this altogether, if it's painful for local dev and we have Lambda timeouts constraining execution in prod anyway.

pkg/service/cognitosvc/service.go
```diff
@@ -0,0 +1,28 @@
...
+
+	"github.com/benbjohnson/clock"
+
+	"github.com/common-fate/ddb"
+	"github.com/common-fate/granted-approvals/pkg/auth"
+	"github.com/common-fate/granted-approvals/pkg/identity"
+	"github.com/common-fate/granted-approvals/pkg/identity/identitysync"
+)
+
+// Service holds business logic relating to Access Requests.
```
chrnorm:
```suggestion
// Service holds business logic relating to Cognito user management.
```

web/src/components/modals/CreateGroupModal.tsx
```diff
@@ -0,0 +1,137 @@
...
+  const methods = useForm<CreateGroupRequestBody>({});
+  const toast = useToast();
+  useEffect(() => {
+    if (!props.isOpen) {
+      methods.reset();
+    }
+  }, [props.isOpen]);
+
+  const onSubmit = async (data: CreateGroupRequestBody) => {
+    console.log({ data });
```
chrnorm:
```suggestion
```

web/src/components/modals/CreateUserModal.tsx
```diff
@@ -13,75 +13,162 @@ import {
...
+  useEffect(() => {
+    if (!props.isOpen) {
+      methods.reset();
+    }
+  }, [props.isOpen]);
 
-  const createUserHandler = () => {
-    setLoading(true);
+  const onSubmit = async (data: CreateUserRequestBody) => {
+    console.log({ data });
```
chrnorm:
```suggestion
```

chrnorm (changes requested):
Looks good, just nitpicks in my comments

chrnorm:
Perhaps we can put a little sync icon in black next to this to indicate it's an action
image

Similar to how we have the "+" icon next to the "Add New" actions
image

pkg/server/routes.go
JoshuaWilkes: Good call, This just slipped in here and my usecase was to do longer debugging

PR #266, opened 2022-09-15
Grant validations 

meyerjrr:


accesshandler/openapi.yml
```diff
@@ -151,6 +151,27 @@ paths:
...
+  "/api/v1/providers/{providerid}/validate":
+    post:
+      summary: ""
+      operationId: validate-request-to-provider
+      responses:
+        "200":
+          $ref: "#/components/responses/HealthResponse"
+        "500":
+          description: Internal Server Error
+      description: Valdates grant will succeed without granting access
```
chrnorm:
```suggestion
      description: Validates grant will succeed without granting access
```

accesshandler/pkg/api/provider.go
```diff
@@ -24,6 +24,34 @@ func (a *API) GetProvider(w http.ResponseWriter, r *http.Request, providerId str
...
+
+	var b types.CreateGrant
+
+	err := apio.DecodeJSONBody(w, r, &b)
+	if err != nil {
+		apio.Error(ctx, w, err)
+		return
+	}
+
+	err = prov.Provider.Validate(ctx, string(b.Subject), []byte{})
```
chrnorm: The third argument here shouldn't be empty - we need to read the request body into a byte array and then send it here. This will cause validation to always fail with providers like `aws-sso`, as they need to read the arguments (like the account ID) in order to validate whether it's possible to grant access.

e.g.

```go
body, err := io.ReadAll(r.Body)
err = prov.Provider.Validate(ctx, string(b.Subject), body)
```

accesshandler/pkg/providers/providers.go
```diff
@@ -18,6 +18,9 @@ type Accessor interface {
 
 	// Revoke the access.
 	Revoke(ctx context.Context, subject string, args []byte, grantID string) error
+
+	// validate the access.
+	Validate(ctx context.Context, subject string, args []byte) error
```
chrnorm: We should remove this from the interface, and instead check whether the provider implements any validation. We already have the `Validator` interface defined below.

e.g. in the new validation API endpoint you've introduced:

```go
prov := config.Providers[providerId]
validator, ok := prov.(providers.Validator)
if !ok {
  // provider doesn't implement validation, so just return a HTTP OK response
}

// the provider implements validation, so try and validate the request
validator.Validate(ctx, subject, args)
```

This way, for simple providers we don't need to implement the Validate() method (as you've had to do with TestVault).

accesshandler/openapi.yml
```diff
@@ -151,6 +151,27 @@ paths:
         name: providerId
         in: path
         required: true
+  "/api/v1/providers/{providerid}/validate":
+    post:
+      summary: ""
+      operationId: validate-request-to-provider
+      responses:
+        "200":
+          $ref: "#/components/responses/HealthResponse"
```
chrnorm: Instead of returning `HealthResponse`, can we return a more informative response here? I'm thinking we can reuse some of the ideas from the Diagnostics responses we wrote for the interactive provider setup flow, including how we have a function which returns validations with different IDs:

```go
ValidateConfig() map[string]providers.ConfigValidationStep 
```

This is nice because it makes debugging a lot easier - it gives us the opportunity to log the diagnostics, and users can send us the output when validation fails.

e.g. returning a `GrantValidationResponse` as per the below:

```yaml
components:
  GrantValidation:
        title: GrantValidation
        type: object
        description: A validation of a Grant, without actually creating it.
        properties:
          id:
            type: string
            description: "The ID of the validation, such as `user-exists-in-sso`."
          logs:
            type: array
            items:
              $ref: "#/components/schemas/DiagnosticLog"
        required:
          - id
          - status
          - fieldsValidated
      DiagnosticLog:
        title: DiagnosticLog
        type: object
        description: A log entry related to a validation.
        properties:
          level:
            type: string
            description: The log level.
            enum:
              - INFO
              - WARNING
              - ERROR
          msg:
            type: string
            description: The log message.
        required:
          - level
          - msg
responses:
   GrantValidationResponse:
        description: A validation of a Grant to verify whether it can be created.
        content:
          application/json:
            schema:
              type: object
              properties:
                validation:
                  $ref: "#/components/schemas/GrantValidation"
              required:
                - validation
```

Then, in the Go code, add a `GrantValidationStep` type, similar to how `ConfigValidationStep` works. Note that we'll need to update the `Validator` interface too - we can reuse the same approach we have taken with `ConfigValidator`:
```diff

+type GrantValidationStep struct {
+	Name            string
+	FieldsValidated []string
+	Run             func(ctx context.Context, subject string, args []byte) diagnostics.Logs
+}


// Validators know how to validate access without actually granting it.
type Validator interface {
	// Validate arguments and a subject for access without actually granting it.
-	Validate(ctx context.Context, subject string, args []byte) error
+	ValidateGrant() map[string] GrantValidationStep
}
```

And then an actual validation function would look something like:

```go
func (p *Provider) ValidateGrant() map[string]GrantValidationStep {
  return map[string]GrantValidationStep{
    "user-exists-in-aws-sso": {
      Name: "The user must exist in the AWS SSO instance",
      Run: func(ctx context.Context, subject string, args []byte) diagnostics.Logs {
        // i've left the actual logic out just for brevity
        err := checkExistsInAWSSSO()
        if err != nil {
          return diagnostics.Error(err)
        }
      }
    }
  }
}
```

chrnorm: This removes the need to do the concurrency stuff inside the Validate method like we are currently doing - instead, we can run it in the API handler, the same way that ValidateConfig works.

pkg/service/accesssvc/create.go
```diff
@@ -68,6 +69,14 @@ func (s *Service) CreateRequest(ctx context.Context, user *identity.User, in typ
 		RuleVersion:     rule.Version,
 		SelectedWith:    make(map[string]access.Option),
 	}
+
+	//validate the request against the access handler - make sure that access will be able to be provisioned
```
chrnorm: Can you please shift this logic to be inside the Granter's `CreateGrant` method, rather than exposed as a separate method?

This current implementation means that in future we could accidentally call `granter.CreateGrant()` without calling `granter.ValidateGrant()` first. If the validation logic sits inside of `granter.CreateGrant()`, we'll always call it.

pkg/service/accesssvc/service.go
```diff
@@ -27,6 +27,7 @@ type Service struct {
 type Granter interface {
 	CreateGrant(ctx context.Context, opts grantsvc.CreateGrantOpts) (*access.Request, error)
 	RevokeGrant(ctx context.Context, opts grantsvc.RevokeGrantOpts) (*access.Request, error)
+	ValidateGrant(ctx context.Context, opts grantsvc.CreateGrantOpts) error
```
chrnorm, line 30: This should be removed - as per my above review comment
```suggestion
```

accesshandler/pkg/providers/testvault/validate.go
```diff
@@ -0,0 +1,11 @@
...
+
+import (
+	"context"
+)
+
+// testvault validate should just run
+func (p *Provider) Validate(ctx context.Context, subject string, args []byte) error {
+
+	return nil
+}
```
chrnorm: This should be removed entirely if the provider doesn't implement any validation (as per my above comments, removing the Validate() method from the Accessor interface will allow us to do this).

chrnorm (changes requested):
Thanks @meyerjrr. My main feedback is that if we borrow some of the ideas from the `ValidateConfig` methods we've added to our providers we can improve the UX a lot here, such as displaying to users/admins which particular validation check failed. We can give them an error like `user-exists-in-aws-sso`, `aws-account-exists`, etc.

pkg/service/grantsvc/granter.go
```diff
@@ -137,6 +137,47 @@ func (g *Granter) RevokeGrant(ctx context.Context, opts RevokeGrantOpts) (*acces
...
+		End:     iso8601.New(end),
+	}
+	for k, v := range opts.AccessRule.Target.With {
+		req.With.AdditionalProperties[k] = v
+	}
+	for k, v := range opts.Request.SelectedWith {
+		req.With.AdditionalProperties[k] = v.Value
+	}
+
+	res, err := g.AHClient.ValidateRequestToProviderWithResponse(ctx, opts.AccessRule.Target.ProviderID, req)
```
JoshuaWilkes: It looks like the response here from validateRequest is discarded.

This means that if there are any validation errors they will be ignored!

I think you will need to expand this to check the contents of the validation response body and report the errors back to the client.

pkg/service/grantsvc/granter.go
```diff
@@ -137,6 +137,47 @@ func (g *Granter) RevokeGrant(ctx context.Context, opts RevokeGrantOpts) (*acces
 	return nil, errors.New("unhandled response code")
 }
 
+// validate grant runs all the checks that will need to occur when creating a real grant to validate its success
+func (g *Granter) ValidateGrant(ctx context.Context, opts CreateGrantOpts) error {
+	q := &storage.GetUser{
```
JoshuaWilkes: I propose that you change the input to this method to avoid needing to call dynamo db again to fetch the user as its already called in the create grant method.

I think you can also make this a private method as it will only be used by createGrant.

Because the types of PostGrantsJSONRequestBody and ValidateRequestToProviderJSONRequestBody are identical, could you use a shared type in the open API schema instead?

That way you can reduce the code duplication, and potentially remove this validation function all together and just have the call to validate as part of create grant

accesshandler/openapi.yml
```diff
@@ -151,6 +151,27 @@ paths:
         name: providerId
         in: path
         required: true
+  "/api/v1/providers/{providerid}/validate":
```
JoshuaWilkes: Why not just make this /api/v1/grants/validate POST?

The Grant request body contains all the information required for the Grant/validation

web/src/pages/access/request/[id].tsx
```diff
@@ -141,8 +143,25 @@ const Home = () => {
     if (data.when === "scheduled") {
       r.timing.startTime = new Date(data.startDateTime).toISOString();
     }
-    await userCreateRequest(r);
-    navigate({ to: "/requests" });
+    await userCreateRequest(r)
```
Comment (JoshuaWilkes) [inline web/src/pages/access/request/[id].tsx:191]: Will this response need to handle more detail if there are validation errors like user not existing?

JoshuaWilkes (changes requested):
Requesting changes to the API/services and also questions about handling when there are validation errors

pkg/service/grantsvc/granter.go
meyerjrr: Good points, Applied the private method one + the user dynamo lookup. I think It will be cleaner to separate concerns with separate functions here though

web/src/pages/access/request/[id].tsx
Comment (meyerjrr) [reply web/src/pages/access/request/[id].tsx:191]: Done, but problem I'm having is how best to display these errors in the frontend. If it is apart of the error response we return a generic 500 error here and all the helpful meaning is lost. Might need some guidance here @JoshuaWilkes

chrnorm (dismissed):
@meyerjrr some UI feedback from me on this one:

Can you please remove the "Grant Validation Test" UI from the frontend? I understand the approach with reusing the existing React component (especially given that our backend checks are now very similar as the provider setup validation). 

However, in this case I feel it adds a lot of unneccessary complexity to the New Request page. Most of the time the checks should all pass, so Granted doesn't need to tell the end user that yes the Permission Set still exists every time.

Additionally from a security perspective we may leak information intended for admin users only through the diagnostic logs.

What I propose instead is:

- If the validation succeeds, we don't show anything in the UI. This should happen in 99% of cases.
- If the validation fails, the API returns a HTTP Bad Request error with the message being "access request validation failed (error code: user-exists-in-aws-sso, account-exists)". These error codes are the IDs of the checks which have failed.
- The frontend displays the error message to the user (and does not display any diagnostic logs)
- The backend logs the full diagnostics so that admins can run `gdeploy logs get` and print the errors

In a future PR we can print error messages into the request audit trail, however as mentioned we should think about restricting the logs so that only admins can see them.

image

meyerjrr:
@chrnorm Updated to display toast errors for validation errors that are propagated through to the frontend. 
And added more descriptive logging in the backend for these validation errors.

If you're happy with this UX I will wrap up the ticket by updating the other provider validation methods to be like the aws-sso-v2 shown in this ticket.

JoshuaWilkes (approved):

PR #270, opened 2022-09-19
added value to dropdown select

meyerjrr:
Adds value to option dropdowns in provider menu
Screen Shot 2022-09-19 at 6 14 51 pm

Ideally wanting to have tooltips here. Will investigate with @jordiup tomorrow
Screen Shot 2022-09-19 at 6 15 04 pm


chrnorm (approved):
Looks good and working for me when I spun it up!

PR #271, opened 2022-09-20
fix: Disable "Approve" button when self-reviewing

alexjurkiewicz:
The button already does nothing and has a tooltip explaining why, but
it's still enabled.


chrnorm:
🙌 thanks for this @alexjurkiewicz!

alexjurkiewicz:
(I didn't test this, btw)

JoshuaWilkes (approved):

PR #278, opened 2022-09-28
Bugfix: failed grant requests can be cancelled 

meyerjrr:
Updates cancel logic to be able to cancel requests when the underlying grant has failed

### DIFF
--- a/pkg/eventhandler/eventhandler.go
+++ b/pkg/eventhandler/eventhandler.go
@@ -82,6 +82,8 @@ func (n *EventHandler) HandleGrantEvent(ctx context.Context, log *zap.SugaredLog
 		}
 		requestEvent = access.NewGrantFailedEvent(gq.Result.ID, event.Time, oldStatus, newStatus, grantFailedEvent.Reason)
 		log.Infow("inserting request event for grant failed")
+		//if failed then we want to update the end time of the grant to be completed
+		gq.Result.Grant.End = event.Time
 	} else {
 		requestEvent = access.NewGrantStatusChangeEvent(gq.Result.ID, event.Time, nil, oldStatus, newStatus)
 		log.Infow("inserting request event for grant status change")
--- a/pkg/eventhandler/eventhandler.go
+++ b/pkg/eventhandler/eventhandler.go
@@ -82,8 +82,7 @@ func (n *EventHandler) HandleGrantEvent(ctx context.Context, log *zap.SugaredLog
 		}
 		requestEvent = access.NewGrantFailedEvent(gq.Result.ID, event.Time, oldStatus, newStatus, grantFailedEvent.Reason)
 		log.Infow("inserting request event for grant failed")
-		//if failed then we want to update the end time of the grant to be completed
-		gq.Result.Grant.End = event.Time
+
 	} else {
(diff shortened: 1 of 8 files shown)


JoshuaWilkes:
@meyerjrr 

I think that the root cause of the request showing up in "upcoming" when the grant has failed is because we don't correctly handle setting the value of the end time that is used to list requests in either upcoming or past.

see this method

```
func (r *Request) DDBKeys() (ddb.Keys, error) {
	// - APPROVED requests have an end time on the grant
	// - PENDING Scheduled requests have a request end time
	// - PENDING asap requests should have MAXIMUM endtime
	// - Declined and Cancelled requests should have an end time = createdAt so they get a somewhat natural order in the results
	// - REVOKED grants should have end time = created at
	end := r.CreatedAt
	if r.Status == APPROVED || r.Status == PENDING {
		if r.Grant != nil {
			if r.Grant.Status != ac_types.GrantStatusREVOKED {
				end = r.Grant.End
			}
		} else if r.IsScheduled() {
			_, end = r.GetInterval()
		} else {
			// maximum time value in Go
			// this means that asap requests which are not approved will always be the first in results because the end time in unknown until approval
			end = time.Unix(1<<63-1, 0)
		}
	}

```

This in in pkg/access/request.go

What should happen here is that if the grant is not nil and the grant status is failed/error it should set the end = r.createdAt this means that the request will move into the "past" column as soon as the grant fails.

This way the UI will show the request as Error and it will be in the past column as expected.

pkg/eventhandler/eventhandler.go
```diff
@@ -82,6 +82,8 @@ func (n *EventHandler) HandleGrantEvent(ctx context.Context, log *zap.SugaredLog
 		}
 		requestEvent = access.NewGrantFailedEvent(gq.Result.ID, event.Time, oldStatus, newStatus, grantFailedEvent.Reason)
 		log.Infow("inserting request event for grant failed")
+		//if failed then we want to update the end time of the grant to be completed
+		gq.Result.Grant.End = event.Time
```
JoshuaWilkes: This is not the right way to do this, the grant start and end time should not be modified after it is created.
These fields indicate the requested grant.

JoshuaWilkes: The specific change that needs to be made is on the DDBKeys method of request.
There is a caculated field there which is used to sort the request into either upcoming or past

JoshuaWilkes (changes requested):

web/src/components/Request.tsx
```diff
@@ -719,11 +719,13 @@ export const RequestCancelButton: React.FC = () => {
   return (
     <ButtonGroup variant="outline" size="sm">
       {!request && <Skeleton rounded="full" w="64px" h="32px" />}
-      {request?.status === "PENDING" && request.grant?.status == "PENDING" && (
-        <Button rounded="full" onClick={handleCancel}>
-          Cancel
-        </Button>
-      )}
+      {/* allow to cancel requests if the grant has not been created yet */}
+      {(request?.status === "PENDING" && request.grant?.status == "PENDING") ||
```
JoshuaWilkes: @meyerjrr the nesting/brackets doesn't look right here, looks like it will only show the cancel button if grant == undefined

chrnorm (approved):

PR #279, opened 2022-09-28
Add okta org URL validation for increased security

JoshuaWilkes:
image

### DIFF
--- a/pkg/service/psetupsvc/complete_step.go
+++ b/pkg/service/psetupsvc/complete_step.go
@@ -66,7 +66,9 @@ func (s *Service) CompleteStep(ctx context.Context, setupID string, stepIndex in
 	}
 
 	// load the current values into this config
-	err = cfg.Load(ctx, &gconfig.MapLoader{Values: setup.ConfigValues})
+	// Skip secrets is set to true because we never need to read the value of a secret in this context.
+	// If a value is provided for a secret it will be updated
+	err = cfg.Load(ctx, &gconfig.MapLoader{Values: setup.ConfigValues, SkipLoadingSecrets: true})
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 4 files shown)


pkg/service/psetupsvc/complete_step.go
```diff
@@ -66,7 +66,9 @@ func (s *Service) CompleteStep(ctx context.Context, setupID string, stepIndex in
 	}
 
 	// load the current values into this config
-	err = cfg.Load(ctx, &gconfig.MapLoader{Values: setup.ConfigValues})
+	// Skip secrets is set to true because we never need to read the value of a secret in this context.
+	// If a value is provided for a secret it will be updated
+	err = cfg.Load(ctx, &gconfig.MapLoader{Values: setup.ConfigValues, SkipLoadingSecrets: true})
```
JoshuaWilkes, line 71: I found a bug in the provider setup UI where config including a secret value failed in teh deployed version because the GetParameter IAM permission is missing from the API server for provider secrets.

What I decided is that in the context of configuration, we never need to retrieve the value of the secret from SSM, so instead, I have added a flag to the config loaders which skips loading secrets as an opt in.

This now works as expected, in the UI you can update any config value without it failing with a permission error, and we don't need to have read access to provider secrets in the client facing API.

pkg/service/psetupsvc/complete_step.go
```diff
@@ -80,7 +82,7 @@ func (s *Service) CompleteStep(ctx context.Context, setupID string, stepIndex in
 
 		// any secret starting with 'awsssm://' is assumed to be an existing reference
 		// to a secret and is not set.
-		if strings.HasPrefix(value, "awsssm://") {
+		if f.IsSecret() && strings.HasPrefix(value, "awsssm://") {
```
JoshuaWilkes, line 85: Adding this additional constraint here as it should really only apply to secret fields

meyerjrr (approved):

PR #280, opened 2022-09-29
Update provider on access rules

JoshuaWilkes:
Allow provider to be updated on an access rule

I updated the Orval config to ensure it prunes old files from the generated api client. This means that when we remove types they are also removed from the frontend files


pkg/api/accessrule.go
```diff
@@ -127,7 +127,7 @@ func (a *API) AdminGetAccessRule(w http.ResponseWriter, r *http.Request, ruleId
 // (POST /api/v1/access-rules/{ruleId})
 func (a *API) AdminUpdateAccessRule(w http.ResponseWriter, r *http.Request, ruleId string) {
 	ctx := r.Context()
-	var updateRequest types.UpdateAccessRuleRequest
+	var updateRequest types.CreateAccessRuleRequest
```
Eddie023, line 130: Is there any reason we are changing `UpdateAccessRuleRequest` to `CreateAccessRuleRequest` . Seems like `UpdateAccessRuleRequest` variable naming was more clear given that we are actually trying to update access rule.

JoshuaWilkes: Because these types are identical I find it makes the typing simpler for the frontend and backend.
If these types diverge again we could switch back to a different type

JoshuaWilkes: Maybe we could make it more generic "AccessRuleRequest"

PR #283, opened 2022-09-30
Fix access rule lookup for cli

JoshuaWilkes:
This PR fixes an issue with the access rule lookup where only the first matching access rule was checked in the backend.

It also make improvements for access rule matches where there are selectable fields for the user. These fields are not pre populated with the correct selections based on the request.

Improved caching in the rule lookup functions to reduce latency

fixed incorrect IDs used during lookups.

Improved the setup documentation for SSO provider to include CLI based deployment of the IAM role


chrnorm (approved):

chrnorm:
Thanks for this @JoshuaWilkes, the tests are much needed here. I ended up refactoring this to be a service method in `rulesvc`, so that changes to our API layer don't break the test cases that you've developed.

JoshuaWilkes:
Good stuff!

PR #284, opened 2022-09-30
CF-145 Add `select all` options to MultiSelect that selects all option

Eddie023:


web/src/components/forms/access-rule/components/Select.tsx
```diff
@@ -138,6 +147,20 @@ export const MultiSelect: React.FC<MultiSelectProps> = ({
             // ref={ref}
             value={options.filter((c) => value.includes(c.value))}
             onChange={(val) => {
+              // for MultiSelect with 'Select All' option
+              // we check if the selected value is 'select all'
+              // if true then add add options as value.
```
Eddie023:
```suggestion
              // if true then add all options as value.
```
```suggestion
              // if true then add all options as value.
```

web/src/components/forms/access-rule/components/Select.tsx
```diff
@@ -112,7 +117,11 @@ export const MultiSelect: React.FC<MultiSelectProps> = ({
             id={id}
             isDisabled={isDisabled}
             //getOptionLabel={(option) => `${option.label}  (${option.value})`}
-            options={options}
+            options={[
```
chrnorm, line 120: @Eddie023 could you make this a button underneath the form field, rather than an option in the list?

Because it's an option, if the user selects it we're going to end up submitting "Select all" to our API alongside the other submitted fields.

chrnorm (changes requested):

web/src/components/forms/access-rule/components/Select.tsx
Eddie023: The `select-all` option is only shown in the frontend and won't be submitted to API.

Eddie023: @chrnorm  This should work as you expect. Here is the loom video https://www.loom.com/share/172c9d95f01c4c6090fddd0f9a7ff340

Also, I personally think having a `select-all` option is a better UX 🤔

chrnorm: @Eddie023 my mistake here - I read through the code without checking out your branch! Thanks for the demo video, it looks really good. Agreed on your approach here. Great work

chrnorm (approved):
Great work @Eddie023!

PR #285, opened 2022-09-30
wait for scheduled start time for instructions

JoshuaWilkes:
This hides the instructions loading animation until the scheduled start time for a scheduled request, it sets teh refresh interval to automatically refresh when the start time passes so you can leave the page running until its ready and then it will load automatically

### DIFF
--- a/web/src/components/Request.tsx
+++ b/web/src/components/Request.tsx
@@ -292,13 +292,31 @@ export const RequestAccessInstructions: React.FC = () => {
 
   useEffect(() => {
     if (reqData?.grant?.status == "PENDING") {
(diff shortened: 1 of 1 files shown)


web/src/components/Request.tsx
```diff
@@ -302,6 +312,14 @@ export const RequestAccessInstructions: React.FC = () => {
     return null;
   }
 
+  // Don't attempt to load a shceduled request until start time
```
chrnorm:
```suggestion
  // Don't attempt to load a scheduled request until start time
```

chrnorm (approved):

PR #274, opened 2022-09-26
feat(web): Use relative timestamps in audit log

alexjurkiewicz:
But preserve the full timestamp as a tooltip.

PR #287, opened 2022-09-30
fix bracket characters in frontend

chrnorm:
Some small frontend UI tidy-ups.

We were displaying `))` in the frontend when access instructions are loading, which has been removed.

Extends on @alexjurkiewicz's work in #274 to show relative timestamps of audit trail entries. I've swapped the `Tooltip` component for a `Popover` to improve accessibility and make it possible to select the extact timestamp text from the frontend UI.

Before:
image

After:
image


alexjurkiewicz:
Nice!

PR #290, opened 2022-10-03
Add waf support to cloudformation stack

JoshuaWilkes:
Add support for WAF firewall ACLS for cloudfront and api gateway.
It is up to the admin to create and configure the WAF rules as they need then update their deployment config following the guide in the docs


JoshuaWilkes:
I have tested with a dev deployment and a gdeploy deployment

chrnorm (approved):
Looks good, we’ll add this to UAT testing prior to the next release

PR #291, opened 2022-10-03
Improve UI for selectable fields

JoshuaWilkes:
image

Improve the UI/UX by using the correct field titles and display all properties being requested


accesshandler/pkg/providers/arg-schemas.go
```diff
@@ -0,0 +1,27 @@
+package providers
+
+import "github.com/invopop/jsonschema"
+
+// GetArgumentTitleFromSchema will lookup an argument in a provider arg schema and return the title if it exists else and empty string
+func GetArgumentTitleFromSchema(s *jsonschema.Schema, argumentID string) string {
```
JoshuaWilkes, line 6: I found this json schema to be imperfect for this usecase, We could consider making the Args/ArgSchema a more rigid type to make this easier

pkg/api/request.go
```diff
@@ -180,8 +183,44 @@ func (a *API) UserGetRequest(w http.ResponseWriter, r *http.Request, requestId s
 		apio.Error(ctx, w, errors.New("access rule result was nil"))
 		return
 	}
+	var options []cache.ProviderOption
```
JoshuaWilkes: I need to look at factoring this out into teh service as it is repeated a few times, 

I will also look at whether I can reduce it to a single DDB query for all the option for a provider rather than individual options

chrnorm:
@JoshuaWilkes just a heads up that @Eddie023 has made a change to the multi select component in #284

PR #295, opened 2022-10-07
CF-171: access rule name bug

jordiup:
Bug fix for:
Access Request titles don't appear in the request history in admin view

### DIFF
--- a/web/src/components/AccessRuleNameCell.tsx
+++ b/web/src/components/AccessRuleNameCell.tsx
@@ -1,20 +1,25 @@
 import { Box, BoxProps, HStack, Text } from "@chakra-ui/react";
 import React from "react";
 import { Link } from "react-location";
+import { useAdminGetAccessRule } from "../utils/backend-client/admin/admin";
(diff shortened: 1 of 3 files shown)


web/src/components/AccessRuleNameCell.tsx
```diff
@@ -1,20 +1,25 @@
...
 import React from "react";
 import { Link } from "react-location";
+import { useAdminGetAccessRule } from "../utils/backend-client/admin/admin";
 import { useUserGetAccessRule } from "../utils/backend-client/end-user/end-user";
 import { ProviderIcon } from "./icons/providerIcon";
 
 type Props = {
   reason?: string;
   accessRuleId: string;
+  adminRoute: boolean;
```
chrnorm, line 11: Ideally we'd make this a presentation-only component that doesn't do any data fetching - this prop couples the way that we are fetching the data with the way we're displaying it. Happy with this approach for now though!

chrnorm (approved):

PR #300, opened 2022-10-10
CF-180: Grant Error Message

jordiup:
UDE = "1 error occurred" doesn't add any info
UDE = "*" character as a dot point makes Granted Approvals feel unpolished

- [x] IO = add a UserErrorMessage field to the GrantValidationStep struct and fill this out for our providers
- [x] IO = update the API to return the UserErrorMessage rather than the Name 

image


JoshuaWilkes (approved):

JoshuaWilkes:
@jordiup Could you please fix the broken test then this can go in

jordiup:
> @jordiup Could you please fix the broken test then this can go in

Cheers Josh, think we're good now

PR #303, opened 2022-10-12
UX Improvements

meyerjrr:
Couple of UX improvments in one branch:
- Fixes table print blocking vision of errors
- Fixes padding around options popover
- Update request time left on interval 
Screen Shot 2022-10-12 at 11 27 43 am

### DIFF
--- a/cmd/gdeploy/commands/status.go
+++ b/cmd/gdeploy/commands/status.go
@@ -18,20 +18,17 @@ var StatusCommand = cli.Command{
 			return err
 		}
 
+		o, err := dc.LoadOutput(ctx)
+		if err != nil {
+			return err
+		}
+		o.PrintTable()
+
 		ss, err := dc.GetStackStatus(ctx)
 
 		if err != nil {
 			return err
 		}
-
-		//wait until we know there are no errors then print the status
-
-		o, err := dc.LoadOutput(ctx)
-
-		if err != nil {
-			return err
-		}
-		o.PrintTable()
 
 		clio.Info("Cloudformation stack status: %s", ss)
 
--- a/cmd/gdeploy/commands/status.go
+++ b/cmd/gdeploy/commands/status.go
@@ -17,18 +17,21 @@ var StatusCommand = cli.Command{
 		if err != nil {
 			return err
 		}
-		o, err := dc.LoadOutput(ctx)
-
-		if err != nil {
-			return err
-		}
-		o.PrintTable()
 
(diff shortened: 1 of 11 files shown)


web/src/pages/requests/index.tsx
```diff
@@ -52,7 +52,9 @@ const Home = () => {
   const search = useSearch<MyLocationGenerics>();
   const navigate = useNavigate<MyLocationGenerics>();
 
-  const { data: rules } = useListUserAccessRules();
+  const { data: rules } = useListUserAccessRules({
```
chrnorm: @meyerjrr is this the right API endpoint to be refreshing? I would expect that we need to refresh the `useUserListRequestsUpcoming` call, rather than this one, in order to automatically refresh the request expiry time.

chrnorm (changes requested):

web/src/pages/requests/index.tsx
meyerjrr: Hmm this seemed to solve the issue in the ticket. I will check with Jordi on the implementation here in the frontend

cmd/gdeploy/commands/status.go
```diff
@@ -17,18 +17,21 @@ var StatusCommand = cli.Command{
...
 
 		if err != nil {
 			return err
 		}
-		o.PrintTable()
 
-		ss, err := dc.GetStackStatus(ctx)
+		//wait until we know there are no errors then print the status
+
+		o, err := dc.LoadOutput(ctx)
```
JoshuaWilkes: @meyerjrr in this case I think we always want it to print the table event if the status is bad.
Otherwise, the user can never access the outputs without going into the AWS console.

JoshuaWilkes (dismissed):

web/src/pages/requests/index.tsx
meyerjrr: Updated @chrnorm, the hooks being passed through seemed to not be working so this has been fixed now.

cmd/gdeploy/commands/status.go
chrnorm: I missed this in my review but yep, the status command should always print a table regardless of the status. Maybe we can print the status after we print the table. The specific user feedback we've got is that the create/update commands print a table, which obscures the actual deployment status.

meyerjrr:
@chrnorm Ready for rereview - change to gdeploy status was undone + added some retry logic for other screens with a request timer.

chrnorm (approved):

PR #304, opened 2022-10-12
added error boundry for unexpected errors

meyerjrr:
Added error boundry to catch unexpected errors and log them in a elegant way
Screen Shot 2022-10-12 at 3 48 32 pm

### DIFF
--- a/web/src/utils/errorBoundry.tsx
+++ b/web/src/utils/errorBoundry.tsx
@@ -1,0 +1,71 @@
+import {
+  Flex,
+  Stack,
+  Heading,
+  Button,
+  Text,
+  Link as ChakraLink,
+} from "@chakra-ui/react";
+import { Link } from "react-location";
+
+import React, { Component, ErrorInfo, ReactNode } from "react";
+import { ApprovalsLogo } from "../components/icons/Logos";
+
+interface Props {
+  children?: ReactNode;
+}
+
+interface State {
+  error: Error | undefined;
+}
+
+class ErrorBoundary extends Component<Props, State> {
+  public state: State = {
+    error: undefined,
+  };
+
+  public static getDerivedStateFromError(error: Error): State {
+    // Update state so the next render will show the fallback UI.
+    return { error: error };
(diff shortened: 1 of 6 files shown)


web/src/utils/errorBoundry.tsx
```diff
@@ -0,0 +1,71 @@
...
+        <>
+          <Flex
+            height="100vh"
+            padding="0"
+            alignItems="center"
+            justifyContent="center"
+          >
+            <Stack textAlign="center" w="70%">
+              <Heading pb="10px">An unexpected error occurred.</Heading>
+              <Text>{this.state.error.message}</Text>
```
Eddie023:
```suggestion
             { this.state.error?.message &&  <Text>{this.state.error.message}</Text>}
```
Every error might not have an associated message with it. If the error object doesn't have message this will through a new error 😂

web/src/utils/errorBoundry.tsx
```diff
@@ -0,0 +1,71 @@
...
+              <Link to={"./requests"}>
```
Eddie023: What about when there is an error in Admin Dashboard? Should we still redirect to "./requests" i.e User View ?

web/src/utils/errorBoundry.tsx
meyerjrr: Good catch @Eddie023 !

web/src/utils/errorBoundry.tsx
```diff
@@ -0,0 +1,71 @@
+import {
```
chrnorm: nitpick: can we rename this to `errorBoundary.tsx`?

chrnorm (dismissed):

Eddie023 (approved):

meyerjrr:
@Eddie023 Refactored the ErrorBoundary into using a HOC to take advantage of `navigate` so we can navigate to the previous page the user was on before running into the error.

JoshuaWilkes (approved):

PR #307, opened 2022-10-13
Add ability to assign dynamic values to access rule arguments. E.g Select all accounts with OrgUnit 'ou-abc'

Eddie023:
1. Create access rule request body is change to:
```
{
	"approval": {
		"users": [],
		"groups": []
	},
	"timeConstraints": {
		"maxDurationSeconds": 3600
	},
	"target": {
		"providerId": "aws-sso-v2",
		"with": {
			"accountId": {
				"values": ["1234566777"],
				"groupings": {
					"organizationalUnit": ["ou-6nde-lolololo"],
					"tag": ["123"]
				}
			},
			"permissionSetArn": {
				"values": ["arn:aws:sso:::permissionSet/"],
				"groupings": {}
			}
		}
	},
	"name": "asd",
	"description": "asd",
	"groups": ["granted_administrators"]
}
```

2. Provider Target has a new field "withDynamicId" where we add stuff that needs to be fetched dynamically when the user request access. 

3. Remove the use of "jsonSchema" and create a custom schema of the type:
```
 ArgSchemaMap{
		"permissionSetArn": {
			ID:          "permissionSetArn",
			Title:       "Permission Set",
			Description: "The AWS Permission Set",
			Type:        "input",
			Filters: map[string]Filter{
				"organizationalUnit": {
					Title: "Organizational Unit",
				},
			},
		},
		"accountId": {
			ID:          "accountId",
			Title:       "Account",
			Description: "The AWS Account ID",
			Type:        "multi-select",
			Filters: map[string]Filter{
				"organizationalUnit": {
					Title: "Organizational Unit",
					Id:    "organizationalUnit",
				},
				"tag": {
					Title: "Tag Name",
					Id:    "tag",
				},
			},
		},
	}

```

5. FE changes to handle new arg schema object. 
6 Add `ArgGroupingView` component to show "filters" "groupings" of the argument.


pkg/rule/accessrule.go
```diff
@@ -215,9 +220,31 @@ func (t Target) ToAPIDetail(argOptions []cache.ProviderOption) types.AccessRuleT
...
+	// 					return at
+	// 				}
+	// 				// TODO: remove duplication
+	// 				t.WithSelectable[arg] = append(t.WithSelectable[arg], moreValues...)
+	// 			}
+
+	// 		}
+	// 	}
+	// }
+
```
Eddie023: This should be removed.

accesshandler/pkg/providers/aws/sso-v2/arg_filters.go
```diff
@@ -0,0 +1,98 @@
...
+			for _, orgUnit := range ou.OrganizationalUnits {
+				opts = append(opts, types.Option{Label: aws.ToString(orgUnit.Name), Value: aws.ToString(orgUnit.Id)})
+			}
+		}
+		return opts, nil
+
+	case "tag":
+		return []types.Option{{Label: "abc", Value: "123"}, {Label: "xyz", Value: "456"}}, nil
+
+	}
```
Eddie023: just want to show that if we need to add a new "filter", the associated filter needs to have a logic to fetch the options here.

accesshandler/pkg/providers/aws/sso-v2/arg_filters.go
```diff
@@ -0,0 +1,98 @@
...
+	return nil, &providers.InvalidFilterIdError{FilterId: filterId}
+
+}
+
+func (p *Provider) FetchArgValuesFromDynamicIds(ctx context.Context, argId string, groupName string, groupValues []string) ([]string, error) {
+	switch argId {
+	case "accountId":
+		switch groupName {
+		case "organizationalUnit":
+
```
Eddie023: Each filter needs to have a way to fetch the actual values here. Need a good approach on how to map the filter values and this logic so that users doesn't forget to add these as well.

web/src/utils/backend-client/admin/admin.ts
```diff
@@ -615,6 +615,51 @@ export const useListProviderArgOptions = <TError = ErrorType<ErrorResponseRespon
...
+    filterId: string,
+ options?: SecondParameter<typeof customInstance>) => {
+      return customInstance<ArgOptionsResponseResponse | void>(
+      {url: `/api/v1/admin/providers/${providerId}/args/${argId}/filters/${filterId}`, method: 'get'
+    },
+      options);
+    }
+  
+
+export const getListProviderArgFiltersKey = (providerId: string,
```
Eddie023: starting to feel like this API call isn't required. And we can add the filter options value to the `listArgOptions` API itself. Thoughts?

pkg/api/accessrule.go
```diff
@@ -266,6 +268,37 @@ func (a *API) UserGetAccessRule(w http.ResponseWriter, r *http.Request, ruleId s
...
+
+					for _, value := range argValueBelongingToGroup {
+						if _, exists := rule.Target.WithSelectable[value]; !exists {
+							rule.Target.WithSelectable[arg] = append(rule.Target.WithSelectable[arg], value)
+						}
+					}
+				}
+			}
+		}
+	}
```
Eddie023: Unhappy with the way I am "injecting" or mutating `rules.Target.WithSelectables` values here if we have `withDynamicId`.

pkg/service/grantsvc/gomock_reflect_2513350472/prog.go
```diff
@@ -0,0 +1,66 @@
+
+package main
+
+import (
+	"encoding/gob"
+	"flag"
+	"fmt"
```
Eddie023: is this file even needed? I think this file can be removed.

accesshandler/pkg/providers/providers.go
```diff
@@ -54,11 +54,26 @@ type ArgSchemarer interface {
 	ArgSchema() *jsonschema.Schema
 }
 
+// TODO: Unable to make this work without interface{}
+// Add correct type here.
+type ArgSchemarerMapper interface {
+	ArgSchemaV2() interface{}
+}
+
```
Eddie023: Need to fix the type here.

accesshandler/pkg/providers/aws/sso-v2/aws_sso.go
```diff
@@ -7,20 +7,23 @@ import (
...
-	client        *ssoadmin.Client
-	idStoreClient *identitystore.Client
-	orgClient     *organizations.Client
-	ssoRoleARN    gconfig.StringValue
-	instanceARN   gconfig.StringValue
+	awsConfig       aws.Config
+	client          *ssoadmin.Client
+	idStoreClient   *identitystore.Client
+	orgClient       *organizations.Client
+	resourcesClient *resourcegroupstaggingapi.Client
```
JoshuaWilkes: Remove the resources client from this PR because it is related to the tags group which we will introduce seperately

accesshandler/pkg/providers/aws/sso-v2/aws_sso.go
```diff
@@ -45,16 +48,44 @@ func (p *Provider) Init(ctx context.Context) error {
 	if err != nil {
 		return err
 	}
+
+	resourcesCfg := cfg.Copy()
```
JoshuaWilkes: As above, don't need this client for this PR

accesshandler/pkg/providers/aws/sso-v2/options.go
```diff
@@ -70,10 +72,7 @@ func (p *Provider) Options(ctx context.Context, arg string) ([]types.Option, err
 					if po.PermissionSet.Name != nil {
 						label = *po.PermissionSet.Name
 					}
-					if po.PermissionSet.Description != nil {
-						label = label + ": " + *po.PermissionSet.Description
-					}
-					opts = append(opts, types.Option{Label: label, Value: ARNCopy})
+					opts.Options = append(opts.Options, types.Option{Label: label, Value: ARNCopy, Description: po.PermissionSet.Description})
```
JoshuaWilkes: Ensure that this description now gets displayed nicely in the Frontend and in Slack messages if required.

It was previously concatenated in the Label. Be sure that the accessrulelookup function still works now that description is not part of the label for SSO roles and the granted CLI

accesshandler/pkg/providers/aws/sso/options.go
```diff
@@ -67,7 +67,7 @@ func (p *Provider) Options(ctx context.Context, arg string) ([]types.Option, err
 						if po.PermissionSet.Description != nil {
 							label = label + ": " + *po.PermissionSet.Description
 						}
-						opts = append(opts, types.Option{Label: label, Value: ARNCopy})
+						opts.Options = append(opts.Options, types.Option{Label: label, Value: ARNCopy})
```
JoshuaWilkes: Update this to so that description is seperated from the label and saved to teh option. See the aws-sso-v2 implementation for reference

accesshandler/pkg/providers/testvault/access.go
```diff
@@ -14,7 +14,7 @@ import (
 )
 
 type Args struct {
-	Vault string `json:"vault" jsonschema:"title=Vault,description=The name of an example vault to grant access to (can be any string),default=demovault"`
+	Vault string `json:"vault"`
```
JoshuaWilkes, line 17: In this instance, a default value was provided in the json schema, I don't recall whether this was used in the FE however it has been excluded from the updated schema.

We could add it back in for Input fields if it hurts UX to not have it

accesshandler/pkg/server/routes.go
```diff
@@ -21,7 +21,7 @@ func (s *Server) Routes() http.Handler {
 	r.Use(chiMiddleware.RequestID)
 	r.Use(chiMiddleware.RealIP)
 	r.Use(chiMiddleware.Recoverer)
-	r.Use(chiMiddleware.Timeout(30 * time.Second))
+	r.Use(chiMiddleware.Timeout(30000 * time.Second))
```
JoshuaWilkes:
```suggestion
	r.Use(chiMiddleware.Timeout(30 * time.Second))
```

pkg/access/request.go
```diff
@@ -160,29 +159,32 @@ func (r *Request) ToAPIDetail(accessRule rule.AccessRule, canReview bool, argOpt
...
-			Value: v.Value,
-			Title: k,
+			Label:             v.Label,
+			Value:             v.Value,
+			Title:             k,
+			FieldDescription:  v.Description,
+			OptionDescription: v.Description,
 		}
 		// attempt to get the title for the argument from the provider arg schema
 		if provider != nil {
```
JoshuaWilkes: Consider wether there is a better option for fetching the provider type which matches a rule.
Also, this should probably be for a specific provider version.

Need to define teh behaviour here e.g, if fields changed between provider versions what would happen and there was no match.

pkg/api/accessrule.go
```diff
@@ -266,6 +266,33 @@ func (a *API) UserGetAccessRule(w http.ResponseWriter, r *http.Request, ruleId s
...
 		return
 	}
+
+	for arg, groupings := range rule.Target.WithArgumentGroupOptions {
+		for group, values := range groupings {
+
+			// if provider arg has values in groupings
+			if len(values) > 0 {
+				for _, value := range values {
+					_, cachedGroup, err := a.Cache.LoadCachedProviderArgGroupOptions(ctx, rule.Target.ProviderID, arg, group, value)
```
JoshuaWilkes: we should be able to make a single call to LoadCachedProviderOptions which includes the arg optiosn and group options

TODO only make a single call to the cache service

pkg/api/accessrule.go
```diff
@@ -266,6 +266,33 @@ func (a *API) UserGetAccessRule(w http.ResponseWriter, r *http.Request, ruleId s
...
+						if !contains(rule.Target.WithSelectable[arg], option) {
+							rule.Target.WithSelectable[arg] = append(rule.Target.WithSelectable[arg], option)
+						}
+					}
+				}
+			}
+		}
+	}
+
 	pq := storage.ListCachedProviderOptions{
```
JoshuaWilkes: Replace this with a call to the cache service as in previous comment

pkg/api/api.go
```diff
@@ -103,8 +103,9 @@ type AccessRuleService interface {
 	UpdateRule(ctx context.Context, in *rulesvc.UpdateOpts) (*rule.AccessRule, error)
 }
 type CacheService interface {
-	RefreshCachedProviderArgOptions(ctx context.Context, providerId string, argId string) (bool, []cache.ProviderOption, error)
-	LoadCachedProviderArgOptions(ctx context.Context, providerId string, argId string) (bool, []cache.ProviderOption, error)
+	RefreshCachedProviderArgOptions(ctx context.Context, providerId string, argId string) (bool, []cache.ProviderOption, []cache.ProviderArgGroupOption, error)
+	LoadCachedProviderArgOptions(ctx context.Context, providerId string, argId string) (bool, []cache.ProviderOption, []cache.ProviderArgGroupOption, error)
+	LoadCachedProviderArgGroupOptions(ctx context.Context, providerId string, argId string, groupId string, groupValue string) (bool, cache.ProviderArgGroupOption, error)
```
JoshuaWilkes, line 109: This may not be required because LoadCachedProvider options includes this result

pkg/api/provider.go
```diff
@@ -98,28 +98,41 @@ func (a *API) GetProviderArgs(w http.ResponseWriter, r *http.Request, providerId
...
-	if res.HasOptions {
-		for _, o := range options {
-			res.Options = append(res.Options, ahTypes.Option{
-				Label: o.Label,
-				Value: o.Value,
-			})
-		}
+
+	for _, o := range options {
+		res.Options = append(res.Options, ahTypes.Option{
```
JoshuaWilkes, line 119: Check whether Description needs to be added here

pkg/api/provider.go
```diff
@@ -98,28 +98,41 @@ func (a *API) GetProviderArgs(w http.ResponseWriter, r *http.Request, providerId
...
+			Label: o.Label,
+			Value: o.Value,
+		})
+	}
+
+	for _, group := range groups {
+		res.Groups.AdditionalProperties[group.Group] = append(res.Groups.AdditionalProperties[group.Group], ahTypes.GroupOption{
```
JoshuaWilkes, line 127: check whether Description needs to be added here

pkg/cache/provider_arg_group.go
```diff
@@ -0,0 +1,30 @@
...
+package cache
+
+import (
+	"github.com/common-fate/ddb"
+	"github.com/common-fate/granted-approvals/pkg/storage/keys"
+)
+
+// ProviderArgGroupOption is an argument option that we've cached
+// from an Access Provider in DynamoDB.
+type ProviderArgGroupOption struct {
```
JoshuaWilkes, line 14: Add optional Description here

pkg/rule/accessrule.go
```diff
@@ -181,103 +186,102 @@ func (a *Approval) IsRequired() bool {
 // I expect this will be different to what gets returned in the api response
 type Target struct {
 	// References the provider's unique ID
-	ProviderID     string              `json:"providerId"  dynamodbav:"providerId"`
-	ProviderType   string              `json:"providerType"  dynamodbav:"providerType"`
-	With           map[string]string   `json:"with"  dynamodbav:"with"`
+	ProviderID   string            `json:"providerId"  dynamodbav:"providerId"`
+	ProviderType string            `json:"providerType"  dynamodbav:"providerType"`
+	With         map[string]string `json:"with"  dynamodbav:"with"`
```
JoshuaWilkes, line 145: TODO: I have considered whether we should deprecate the "with" field on this struct and migrate to only using withSelectable and withArgumentGroupOptions

Or a single field where the group options are nested on a struct with each argument, the same as the CreateAccessRuleTarget type.

For the deprecation, we may be able to add some code to the DDB query for AccessRule which converts any Rules at read time from the old schema to the new schema.

pkg/server/routes.go
```diff
@@ -18,7 +18,7 @@ func (c *Server) Handler() http.Handler {
 	r.Use(c.requestIDMiddleware)
 	r.Use(chiMiddleware.RealIP)
 	r.Use(chiMiddleware.Recoverer)
-	r.Use(chiMiddleware.Timeout(30 * time.Second))
+	r.Use(chiMiddleware.Timeout(30000 * time.Second))
```
JoshuaWilkes:
```suggestion
	r.Use(chiMiddleware.Timeout(30 * time.Second))
```

pkg/service/accesssvc/create.go
```diff
@@ -49,6 +49,29 @@ func (s *Service) CreateRequest(ctx context.Context, user *identity.User, in typ
...
 
 	now := s.Clock.Now()
+
+	for arg, groupings := range rule.Target.WithArgumentGroupOptions {
+		for group, values := range groupings {
+
+			// if provider arg has values in groupings
+			if len(values) > 0 {
+				for _, value := range values {
+					_, cachedGroup, err := s.Cache.LoadCachedProviderArgGroupOptions(ctx, rule.Target.ProviderID, arg, group, value)
```
JoshuaWilkes: Consider querying for every argument for the provider rather than querying for every argument. to save on ddb queries/latency

pkg/service/cachesvc/cache.go
```diff
@@ -79,34 +96,65 @@ func (s *Service) RefreshCachedProviderArgOptions(ctx context.Context, providerI
...
+		return ahtypes.ArgOptionsResponse{}, errors.New(*res.JSON500.Error)
 	default:
 		logger.Get(ctx).Errorw("unhandled access handler response", "response", string(res.Body))
-		return types.ArgOptionsResponse{}, fmt.Errorf("unhandled response code: %d", code)
+		return ahtypes.ArgOptionsResponse{}, fmt.Errorf("unhandled response code: %d", code)
 	}
 }
+
+// Load cached provider argument group option's option.
+func (s *Service) LoadCachedProviderArgGroupOptions(ctx context.Context, providerId string, argId string, groupId string, groupValue string) (bool, cache.ProviderArgGroupOption, error) {
```
JoshuaWilkes, line 149: I think this shoudl be removed because it does not offer any new behaviour to the other cache api and also does not refresh the cache

JoshuaWilkes (dismissed):

accesshandler/pkg/providers/testvault/access.go
Eddie023: @JoshuaWilkes It was used as a placeholder in the FE.  As a quick fix; I added `placeholder={`default-${argument.title}`}

pkg/service/cachesvc/cache.go
Eddie023: This only made sense if we wanted to fetch the dynamic values for specific group options only. This could be deleted if we opt to fetch everything related to arg in one single API.

PR #308, opened 2022-10-13
Deprecate aws sso

meyerjrr:
Adds field to enable the ability to deprecate providers
Add logic in the frontend to gracefully not allow users to use a deprecated provider
Remove permissions from access handler for over privileged sso provider implementation 
Screen Shot 2022-10-13 at 3 00 23 pm


accesshandler/pkg/config/providers.go
```diff
@@ -17,16 +17,18 @@ import (
...
 
 type Provider struct {
-	ID       string
-	Type     string
-	Version  string
-	Provider providers.Accessor `json:"-"`
+	ID         string
+	Type       string
+	Version    string
+	Deprecated bool
```
Eddie023: nitpick: Having `is` prefix to `bool` variables is usually considered a good practice. 
```suggestion
	IsDepricated bool
```

Eddie023 (approved):

JoshuaWilkes:
To be honest I think we could just remove the aws-sso v1 provider all together.
Or at minimum, require it be removed from the deployment config before the deployment can be updated to the new version.

We need to ensure that we support deleting an access provider which has been used to create access rules and requests in the past. Could you run some checks on this before we bring this in?

meyerjrr:
> To be honest I think we could just remove the aws-sso v1 provider all together. Or at minimum, require it be removed from the deployment config before the deployment can be updated to the new version.
> 
> We need to ensure that we support deleting an access provider which has been used to create access rules and requests in the past. Could you run some checks on this before we bring this in?

Hmm my angle here was to not totally drop the support for v1, in preference of not introducing a breaking change to the configuration of the app.  Does just removing it kind of defeat the purpose of having versioned providers?
Although removing it from the codebase, would simplify things, and mitigate implementing features in the the wrong place.

I say we could include some encouraging messaging in the app to get users to remove the sso v1 from their config.

chrnorm (changes requested):
Just marking as request changes so we can address the comments raised by @JoshuaWilkes

chrnorm:
> Does just removing it kind of defeat the purpose of having versioned providers?

@meyerjrr good question here, I see the provider versioning being more useful in future when we support community access providers. These providers will be versioned with tags in GitHub repositories, similar to how GitHub Actions are versioned.

meyerjrr:
@JoshuaWilkes @chrnorm 
I had a play around with workflows today and wanted some thoughts on how the UX should play out with a *deprecated* provider / version.

Assuming we just remove the v1 version of aws sso. and require users to remove it from their deployment config.

1. When access rules for a deprecated version exist in a deployment, what should happen when users request access?
- Filter out access rules that are using a deprecated version (dont display them in the frontend, this would need some additional config like `isDeprecated` field)
- Add validation in request access API's to fail when attempting to request access to a rule with a deprecated/removed version
- The rules still exist but they will fall back to another version, in this case will attempt to create the request with aws-sso-v2. (If this is possible)

Keen to hear your thoughts.

chrnorm:
FYI I tried deleting the AWS SSO v1 provider from an internal deployment, and users can't make requests to the access rule anymore. I think this behaviour is good, a lot better than having the access request go through and then it failing at the provisioning stage.

image

chrnorm:
> When access rules for a deprecated version exist in a deployment, what should happen when users request access?

I think we should go with this option: add validation in request access API's to fail when attempting to request access to a rule with a deprecated/removed version.

We may already have this behaviour implemented. What happens if you do the following?

1. Create an Access Rule with `aws-sso@v1`
2. Delete the `aws-sso@v1` provider entirely from the codebase
3. Redeploy Granted Approvals
4. Try and make an access request to the rule

Ideally we want to fail here with a 400 Bad Request error, rather than an internal server error.

meyerjrr:
Tested and default functionality when the provider is removed is to return with the "provider not found" error as shown above.

 
Screen Shot 2022-10-20 at 11 01 18 am
Although it will be a breaking change to bring in and we might want to inform users of this change.

JoshuaWilkes (approved):

chrnorm (dismissed):
Merge conflicts need resolving here

PR #313, opened 2022-10-17
APP-128 Create API in Access Handler to get dynamic group values

Eddie023:


pkg/service/accesssvc/create.go
```diff
@@ -49,6 +50,33 @@ func (s *Service) CreateRequest(ctx context.Context, user *identity.User, in typ
...
+						rule.Target.WithSelectable[arg] = append(rule.Target.WithSelectable[arg], value)
+					}
+				}
+
+			}
+		}
+	}
+
+	// TODO: here
+	// fetch all the dynamic values and add them to withSelectables.
```
Eddie023, line 78: remove this

JoshuaWilkes:
@Eddie023 Could you please make this also take the groupID

So you can fetch the values for a single group value

fetch all accounts for OU 1234

You may also find it convenient to add an api method which just validates an account value for a group is valid

e.g /api/v1/providers/{providerId}/args/{argId}/groups/{groupID}/values/{valueID}

JoshuaWilkes (approved):

PR #325, opened 2022-10-20
hotfix: regex update for required|optional fields

jordiup:
* Allow empty strings as input on non-required fields


Makefile
```diff
@@ -9,5 +9,5 @@ gdeploy:
 generate:
 	go generate ./...
 	cd web && pnpm clienttypegen 
-	pnpm prettier  --write **/openapi.yml
+	pnpm prettier  --write **/**/openapi.yml
```
JoshuaWilkes: @jordiup is this intentional?
did you mean to add this to target the accesshandler/openapi.yml file?

maybe you need both of these

JoshuaWilkes: Is this intentional?

jordiup: Hmmmm, I was noticing a git diff on one of the openapi files (which was untouched), so I thought this would work as a better catch-all for prettier
image
Seems like the current syntax works fine, perhaps the styling just wasn't committed 🤷 
Will revert

JoshuaWilkes (approved):

PR #337, opened 2022-10-26
updated overlap logic for creating new grants

meyerjrr:
Current logic was not taking current active grants into account when checking for overlap. Allowing duplicate ASAP grants to be created.


JoshuaWilkes:
UAT I found some critical issues with this ticket

- [ ] I created a request then another for the same arguments right away, I got internal server error in the FE and this log in the BE 

```
{"level":"error","ts":1667271763.811933,"caller":"apio/response.go:65","msg":"web handler error","reqId":"joshuas-MBP.hub/9gm6DEHCZ9-000068","error":"this request overlaps an existing grant"}
```
image


- [ ] I created a request then another for the same rule but different arguments, I got an internal server error with the same this request overlaps existing grant error log

JoshuaWilkes (dismissed):
See comments

meyerjrr:
@JoshuaWilkes just needing some more info on your findings:

> UAT I found some critical issues with this ticket
> 
> * [ ]  I created a request then another for the same arguments right away, I got internal server error in the FE and this log in the BE
> 
Is the correct DE? 
> ```
> {"level":"error","ts":1667271763.811933,"caller":"apio/response.go:65","msg":"web handler error","reqId":"joshuas-MBP.hub/9gm6DEHCZ9-000068","error":"this request overlaps an existing grant"}
> ```
> 
> image
> 
> * [ ]  I created a request then another for the same rule but different arguments, I got an internal server error with the same this request overlaps existing grant error log
- Were the arguments here for non overlapping timings? If so this would be correct I think

chrnorm:
quick note on my side RE your queries @meyerjrr - if the error is intended behavior we should make it return a HTTP Bad Request response rather than an Internal Server Error - the 500 internal server error makes it seem like things aren’t working properly. Will leave with Josh to clarify on behaviour

pkg/api/request.go
```diff
@@ -232,6 +232,8 @@ func (a *API) UserCreateRequest(w http.ResponseWriter, r *http.Request) {
 		err = apio.NewRequestError(err, http.StatusUnauthorized)
 	} else if err == accesssvc.ErrRuleNotFound {
 		err = apio.NewRequestError(fmt.Errorf("access rule %s not found", incomingRequest.AccessRuleId), http.StatusNotFound)
+	} else if err == accesssvc.ErrRequestOverlapsExistingGrant {
+		err = apio.NewRequestError(fmt.Errorf(err.Error()), http.StatusBadRequest)
```
chrnorm: fmt.Errorf isn't needed here

```suggestion
		err = apio.NewRequestError(err.Error(), http.StatusBadRequest)
```

pkg/service/accesssvc/add_review.go
```diff
@@ -136,20 +128,109 @@ func (s *Service) AddReviewAndGrantAccess(ctx context.Context, opts AddReviewOpt
...
+
+	currentRequestArguments := make(map[string]string)
+	for k, v := range currentRequestRule.Target.With {
+		currentRequestArguments[k] = v
+	}
+	for k, v := range req.SelectedWith {
+		currentRequestArguments[k] = v.Value
+	}
+
+	//todo move time checking overlap into its own function
```
chrnorm:
```suggestion
```

chrnorm (approved):
Looks good to me! I have shifted the overlap checking into it's own file as I had trouble finding it when navigating through the repo, and I've added an additional test case which matched the behaviour I expected (the case I added was `overlap where one rule has selectable with, one rule has regular with`). Great work @meyerjrr

PR #338, opened 2022-10-28
Request multiple target combinations, Save requests as favorites

JoshuaWilkes:
This Draft PR implements a basic POC for how request bookmarks could work and how requests users can request multiple targets at the same time without needing to create multiple requests.

The combination proposed allows users to take control of their most productive workflows and have a repeatable experience with the least number of clicks.

I have opted to avoid making changes to the way requests work other than on the create request form.

This POC does not include any tests and some API behaviour and validation is not yet included.
The frontend UI is MVP for the feature to work and is subject to design feedback


openapi.yml
```diff
@@ -1628,7 +1677,16 @@ components:
         id: 8oa3c4w7ml7tx
       type: object
       additionalProperties:
-        type: string
+        type: array
+        items:
+          type: string
+    CreateRequestWithSubRequest:
```
chrnorm, line 1683: nitpick: CreateRequestGroup might be more concise here

```suggestion
    CreateRequestGroup:
```

pkg/api/favorites.go
```diff
@@ -0,0 +1,76 @@
...
+
+	"github.com/common-fate/apikit/apio"
+	"github.com/common-fate/ddb"
+	"github.com/common-fate/granted-approvals/pkg/auth"
+	"github.com/common-fate/granted-approvals/pkg/storage"
+	"github.com/common-fate/granted-approvals/pkg/types"
+)
+
+// (GET /api/v1/favorites)
+func (a *API) UserListFavorites(w http.ResponseWriter, r *http.Request) {
```
chrnorm, line 16: We need a delete endpoint here to get rid of a favorite a user creates accidentally

chrnorm (changes requested):
- [x] Minor styling issue I found in the preview deployment. This should be left-aligned rather than center-aligned.

image

- [ ] also we need a way for a user to delete a favorite.

- [x] some other UX feedback - rather than showing the empty state UI, can we not show the "Favorites" section at all on the dashboard if it's empty? (it takes up space and pushes the rules list down)

- [x] Finally some minor spacing feedback

image
image

chrnorm:
Other than this the implementation is looking fantastic

chrnorm (dismissed):
@JoshuaWilkes last requested change on this is the ability to delete a favorite. 

DE = if the request is already favorited, clicking the star button should show a modal asking the user to confirm deleting the favorite. If they click yes it is removed

image

PR #340, opened 2022-10-31
Make identity sync safe for go routines

meyerjrr:
Added locking to auth middleware to stop multiple users being saved to the db

### DIFF
--- a/pkg/auth/auth.go
+++ b/pkg/auth/auth.go
@@ -52,6 +52,8 @@ type IdentitySyncer interface {
 // connected identity provider.
 func Middleware(authenticator Authenticator, db ddb.Storage, idp IdentitySyncer) func(next http.Handler) http.Handler {
 	return func(next http.Handler) http.Handler {
+		var isLocked bool
+
 		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 			ctx := r.Context()
 			log := logger.Get(ctx)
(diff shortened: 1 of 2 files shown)


pkg/auth/auth.go
```diff
@@ -77,11 +79,12 @@ func Middleware(authenticator Authenticator, db ddb.Storage, idp IdentitySyncer)
...
-			// which we can use for handling IDP webhook notifications rather than polling.
-			if err == ddb.ErrNoItems {
+			// // if we get ddb.ErrNoItems, the user may not have been synced yet from the IDP.
+			// // try and sync them now.
+			// // Note: this approach isn't very performant. In future this can be replaced with an incremental sync,
+			// // which we can use for handling IDP webhook notifications rather than polling.
+			if err == ddb.ErrNoItems && !isLocked {
+				isLocked = true
 				log.Info("user does not exist in database - running an IDP sync and trying again", "user", claims)
 				err = idp.Sync(ctx)
```
JoshuaWilkes: @meyerjrr I'm wondering if this would be better implemented using a mutex in the idp.Sync(ctx) method

JoshuaWilkes (approved):
@meyerjrr I have changed the implementation to be a mutex in the IDP sync. This has the same effect but will work everywhere we use the IDP sync.

Great work on identifying the cause of this issue!

PR #341, opened 2022-10-31
Enable KMS key rotation [APP-174]

sbldevnet:
Enable KMS key automatic rotation. 
This is required by many security standards and it is a good practice.


chrnorm:
🙌 thanks so much for the contribution @sbldevnet!

chrnorm (approved):

PR #342, opened 2022-11-01
Internal groups

meyerjrr:
Update Group API's to include param for status
Add concept of internal granted groups


openapi.yml
```diff
@@ -1422,13 +1426,14 @@ components:
...
     IdpStatus:
+      type: string
       title: IdpStatus
       x-stoplight:
         id: a3fdr17gch04a
-      type: string
       enum:
         - ARCHIVED
         - ACTIVE
+        - INTERNAL
```
chrnorm: @meyerjrr can we expose this as a separate field called "source", rather than adding it to the status field?

the source field would be an enum with "INTERNAL" and "EXTERNAL" being the options.

chrnorm: This is to deal with the fact that internal groups can also be archived

pkg/storage/list_active_groups.go
```diff
@@ -0,0 +1,33 @@
...
+
+	//Filter expression to not pull internal groups.
+	var expr string
+
+	expr += "#group_status <> :key"
+	key := "ARCHIVED"
+	qi.ExpressionAttributeValues[":key"] = &types.AttributeValueMemberS{Value: key}
+	qi.ExpressionAttributeNames["#group_status"] = "status"
+	qi.FilterExpression = &expr
+	return &qi, nil
```
chrnorm: Why not include all groups here? I would expect ListActiveGroups to return all of the active internal and external groups

openapi.yml
chrnorm: it would be even better actually if we could have the source be the IDP type, so our sources can be:

- internal
- azure
- google
- onelogin

etc

chrnorm (dismissed):
Good start on this @meyerjrr, some initial feedback from me here!

pkg/storage/list_groups_for_source.go
```diff
@@ -0,0 +1,33 @@
...
+		KeyConditionExpression: aws.String("PK = :pk1"),
+		ExpressionAttributeValues: map[string]types.AttributeValue{
+			":pk1": &types.AttributeValueMemberS{Value: keys.Groups.PK1},
+		},
+		ExpressionAttributeNames: make(map[string]string),
+	}
+
+	var expr string
+
+	expr += "#group_source = :key"
```
JoshuaWilkes: Can you make this use a GSI instead so that you can do a simple query without the filter

pkg/service/grantsvc/gomock_reflect_3692639278/prog.go
```diff
@@ -0,0 +1,66 @@
+
+package main
```
JoshuaWilkes: Remove this file

pkg/identity/identitysync/cognito.go
```diff
@@ -67,6 +67,7 @@ func groupFromCognitoGroup(cognitoGroup types.GroupType) identity.IDPGroup {
 		ID:          aws.ToString(cognitoGroup.GroupName),
 		Name:        aws.ToString(cognitoGroup.GroupName),
 		Description: aws.ToString(cognitoGroup.Description),
+		Source:      "COGNITO",
```
JoshuaWilkes: Looks like you would need to add a source for all the other IDP types

pkg/api/groups.go
```diff
@@ -71,25 +114,61 @@ func (a *API) GetGroup(w http.ResponseWriter, r *http.Request, groupId string) {
...
+	// 	apio.JSON(ctx, w, group.ToAPI(), http.StatusCreated)
+	// } else {
+	// Create internal group
+	group := identity.Group{
+		ID:          createGroupRequest.Name,
+		IdpID:       createGroupRequest.Name,
 		Name:        createGroupRequest.Name,
-		Description: aws.ToString(createGroupRequest.Description),
-	})
+		Description: *createGroupRequest.Description,
```
JoshuaWilkes: You will need to check that description is not nil here otherwise you may get a nil pointer exception

pkg/api/groups.go
```diff
@@ -71,25 +114,61 @@ func (a *API) GetGroup(w http.ResponseWriter, r *http.Request, groupId string) {
...
+		Status:      types.IdpStatusACTIVE,
+		Source:      types.INTERNAL,
+		Users:       *createGroupRequest.Members,
```
JoshuaWilkes: If members is not optional and should instead be an empty list. Update the API schema so that this is not a pointer to a slice.

pkg/api/groups.go
```diff
@@ -71,25 +114,61 @@ func (a *API) GetGroup(w http.ResponseWriter, r *http.Request, groupId string) {
...
-		return
-	}
 	var createGroupRequest types.CreateGroupJSONRequestBody
 	err := apio.DecodeJSONBody(w, r, &createGroupRequest)
 	if err != nil {
 		apio.Error(ctx, w, apio.NewRequestError(err, http.StatusBadRequest))
 		return
 	}
-	group, err := a.Cognito.CreateGroup(ctx, cognitosvc.CreateGroupOpts{
+	// if a.Cognito == nil {
```
JoshuaWilkes: Remove this if it is no longer in use

JoshuaWilkes (changes requested):

JoshuaWilkes:
- [x] external groups should have a read only styled form
- [x] fix the background colors
- [x] dont show the cross symbol on the chips for members
image

improve styling on edit group form for internal group

- [x] add submitting state when clicking save
- [x] add form validation
- [x] show a toast when a group is updated

image
Use the same elements as the user form 
image
image

openapi.yml
```diff
@@ -698,6 +698,25 @@ paths:
         $ref: "#/components/requestBodies/CreateGroupRequest"
       tags:
         - Admin
+  /api/v1/admin/groups/source:
```
JoshuaWilkes: Remove this endpoint

Instead, just add a query parameter to the existing list-groups endpoint
This should be an enum with only 2 values "INTERNAL" "EXTERNAL"

pkg/api/groups.go
```diff
@@ -1,19 +1,18 @@
...
 	"github.com/common-fate/granted-approvals/pkg/identity"
-	"github.com/common-fate/granted-approvals/pkg/service/cognitosvc"
 	"github.com/common-fate/granted-approvals/pkg/storage"
 	"github.com/common-fate/granted-approvals/pkg/types"
 )
 
-// Your GET endpoint
+// Gets all active groups
 // (GET /api/v1/groups/)
 func (a *API) GetGroups(w http.ResponseWriter, r *http.Request, params types.GetGroupsParams) {
```
JoshuaWilkes: Could you please rename these existing list endpoints from "GetGroups" -> "ListGroups"

pkg/identity/identitysync/sync.go
```diff
@@ -18,16 +18,16 @@ import (
 )
 
 type IdentityProvider interface {
-	Name() string
```
JoshuaWilkes, line 21: @chrnorm FYI we removed this from the interface and instead get this context from the IdentitySyncer struct after initialisation.

This makes it simpler to implement groups source and avoids a class of issue where groups returned by an idp have different sources or we use the wrong idptype IDs in the code.

We found that the Name() methods were returning different IDs than the IDPType constants. Was this intentional? and is it a breaking change for analytics?

chrnorm: Not a breaking change for analytics! The inconsistency here was just my personal preference (in hindsight I preferred “onelogin” instead of “one-login”, “azure-ad” instead of “azure”). Happy to change these back to be consistent with the YAML config to keep things simple and consistent - so please remove the inconsistencies I’ve introduced.

pkg/api/groups.go
```diff
@@ -57,39 +75,77 @@ func (a *API) GetGroup(w http.ResponseWriter, r *http.Request, groupId string) {
...
+			return
+		}
+		u.Result.Groups = append(u.Result.Groups, group.ID)
+		items = append(items, u.Result)
+	}
+
+	err = a.DB.PutBatch(ctx, items...)
 	if err != nil {
-		apio.Error(ctx, w, err)
+		apio.Error(ctx, w, apio.NewRequestError(err, http.StatusBadRequest))
```
JoshuaWilkes, line 120: A ddb error here would be an internal server error
```suggestion
		apio.Error(ctx, w, err)
```

pkg/api/groups.go
```diff
@@ -57,39 +75,77 @@ func (a *API) GetGroup(w http.ResponseWriter, r *http.Request, groupId string) {
...
+		//update existing group
+		group = identity.Group{
+			ID:          *createGroupRequest.Id,
+			IdpID:       createGroupRequest.Name,
+			Name:        createGroupRequest.Name,
+			Description: aws.ToString(createGroupRequest.Description),
+			Status:      types.IdpStatusACTIVE,
+			Source:      identity.INTERNAL,
+			Users:       createGroupRequest.Members,
+			CreatedAt:   time.Now(),
```
JoshuaWilkes: created at should not change on an update

I would also prefer to see these as a put and post operate seperately

JoshuaWilkes (changes requested):

JoshuaWilkes (commented):
Currently we need to update the User API to allow internal groups to be assigned to a user in the frontend.

We also need to consider what the frontend UI allows when cognito is not the IDP, i think it currently hides the update user modal but we will probably want to include it for internal groups

JoshuaWilkes (approved):

PR #345, opened 2022-11-03
Resolve performance issue with SSOv2 list arg options method [APP-178]

JoshuaWilkes:
This resolves an issue encountered by some users with large AWS organizations for example ~100 accounts.

In some cases when the ListOrganizationalUnitsForParent api call does not return results on the first request, the function would enter an infinite loop until terminated by an API timeout because the nextToken was not being included in the request.

### DIFF
--- a/accesshandler/pkg/server/routes.go
+++ b/accesshandler/pkg/server/routes.go
@@ -2,7 +2,6 @@ package server
 
 import (
 	"net/http"
-	"time"
 
 	"github.com/common-fate/apikit/logger"
 	"github.com/common-fate/apikit/openapi"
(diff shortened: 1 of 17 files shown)


accesshandler/pkg/server/routes.go
```diff
@@ -21,7 +20,6 @@ func (s *Server) Routes() http.Handler {
 	r.Use(chiMiddleware.RequestID)
 	r.Use(chiMiddleware.RealIP)
 	r.Use(chiMiddleware.Recoverer)
-	r.Use(chiMiddleware.Timeout(30 * time.Second))
```
meyerjrr, line 24: Is this not needed because the lambda timeout supersedes this one?

meyerjrr (approved):

accesshandler/pkg/server/routes.go
JoshuaWilkes: yes

PR #347, opened 2022-11-04
regex validation adjustment [APP-180]

jordiup:


web/src/pages/access/request/[id].tsx
```diff
@@ -389,12 +389,12 @@ const Home = () => {
                       {...register("reason", {
                         validate: (value) => {
                           const res: string[] = [];
```
Comment (JoshuaWilkes) [inline web/src/pages/access/request/[id].tsx]: Just use the built in maxLength validation for this rather than the validate function

JoshuaWilkes (changes requested):

JoshuaWilkes:
@jordiup could you please merge with main and regenerate the api

PR #353, opened 2022-11-09
Update google idpsetup

meyerjrr:
Adds override function on Field type which allows for inputting a file path as opposed to a value string.

### DIFF
--- a/pkg/gconfig/gconfig.go
+++ b/pkg/gconfig/gconfig.go
@@ -319,6 +319,7 @@ func WithDefaultFunc(df func() string) FieldOptFunc {
 	}
 }
 
+// WithCLIPrompt allows to override the type of cli prompt asked when setting up an IDP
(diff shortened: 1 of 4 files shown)


pkg/gconfig/gconfig.go
```diff
@@ -306,6 +319,13 @@ func WithDefaultFunc(df func() string) FieldOptFunc {
 	}
 }
 
+// WithCLIPrompt allows to override the type of cli prompt asked when setting up an IDP
```
JoshuaWilkes:
```suggestion
// WithCLIPrompt allows to override the type of cli prompt used to collect a value for this field when used in the context of a CLI
```

JoshuaWilkes (changes requested):

JoshuaWilkes (approved):

PR #356, opened 2022-11-10
Add DeprecatedWarn to gdeploy providers

jordiup:


JoshuaWilkes:
Only the interactive add command workflow is deprecated.
if a user runs `gdeploy provider add` and does not pass a value for the `with` flag, then you should display a prompt telling then that it is deprecated and that we recommend they use the interactive setup in the web app.

`&cli.StringSliceFlag{Name: "with", Usage: "Configuration settings for the provider, in key=value pairs"},`

```
if len(c.StringSlice("with")) == 0 {
 promptDeprecated()
}

```

cmd/gdeploy/commands/provider/provider.go
```diff
@@ -10,5 +14,35 @@ var Command = cli.Command{
...
+	Action: cli.ActionFunc(func(ctx *cli.Context) error {
+		cli.ShowSubcommandHelp(ctx)
+		DeprecatedWarn(ctx)
+		return nil
+	}),
+}
+
+func DeprecatedWarn(c *cli.Context) error {
+	// add a deprecated warning here
+	clio.Warn("Warning: this command is deprecated and no longer supported")
```
JoshuaWilkes: I think we should just redirect the user and explain the benefits. We will continue to support the add command but the detailed instructions will be available in the web UI.

```suggestion
	clio.Warn("Configuring access providers via gdeploy has been deprecated in favour of an interactive setup flow in your admin dashboard. Head there now get started setting up your provider.")
```

cmd/gdeploy/commands/provider/remove.go
```diff
@@ -17,6 +17,9 @@ var removeCommand = cli.Command{
 	},
 	Action: func(c *cli.Context) error {
 		ctx := c.Context
+
+		DeprecatedWarn(c)
```
JoshuaWilkes:
```suggestion
```

cmd/gdeploy/commands/provider/update.go
```diff
@@ -22,6 +22,9 @@ var updateCommand = cli.Command{
 	},
 	Action: func(c *cli.Context) error {
 		ctx := c.Context
+
+		DeprecatedWarn(c)
```
JoshuaWilkes:
```suggestion
```

cmd/gdeploy/commands/provider/provider.go
```diff
@@ -10,5 +14,35 @@ var Command = cli.Command{
 	Subcommands: []*cli.Command{
 		&addCommand, &removeCommand, &updateCommand,
 	},
-	Action: cli.ShowSubcommandHelp,
+	Action: cli.ActionFunc(func(ctx *cli.Context) error {
+		cli.ShowSubcommandHelp(ctx)
+		DeprecatedWarn(ctx)
```
JoshuaWilkes: Only show this on the add command, and only when no "with" flag arguments are provided
```suggestion
```

JoshuaWilkes (changes requested):

JoshuaWilkes (approved):

PR #372, opened 2022-11-21
Internal group tests

meyerjrr:
Create internal group
Update the name description and members
create the rule, then assert that the user has access to the rule and that the admin is required to review it
Admin can manage internal groups for a User


web/package.json
```diff
@@ -30,6 +30,7 @@
     "axios": "^0.27.2",
     "date-fns": "^2.28.0",
     "framer-motion": "^6.3.4",
+    "playwright": "^1.28.0",
```
JoshuaWilkes: be sure not to include playwright as a standard dependency because it's already included as a dev dependency

web/playwright.config.ts
```diff
@@ -1,15 +1,15 @@
...
   forbidOnly: !!process.env.CI,
-  
- 
+
   retries: 0,
   globalSetup: "./globalSetup.ts",
   use: {
     trace: "on",
     baseURL: process.env.TESTING_DOMAIN,
+    headless: false,
```
JoshuaWilkes:
```suggestion
```

web/tsconfig.json
```diff
@@ -16,5 +16,5 @@
     "jsx": "preserve",
     "types": ["vite/client"]
   },
-  "include": ["./src"]
+  "include": ["./src", "."]
```
JoshuaWilkes, line 19: You shouldn't need this as the original setting just forces our prod code to be checked rather than the other code.
Was this to solve some issue locally for you?

JoshuaWilkes (changes requested):

web/tsconfig.json
meyerjrr: Yeah without that I was getting import errors in playwright config

web/src/components/tables/GroupsTable.tsx
```diff
@@ -127,6 +128,13 @@ export const GroupsTable = () => {
         emptyText: "No groups",
         apiPaginator: paginator,
         linkTo: true,
+        rowProps: (row) => ({
+          "_hover": { bg: "gray.50" },
+          "cursor": "pointer",
+          // in our test cases we use reason for the unique key
```
JoshuaWilkes:
```suggestion
```

web/src/components/tables/UserReviewsTable.tsx
```diff
@@ -128,7 +128,7 @@ export const UserReviewsTable = () => {
           "_hover": { bg: "gray.50" },
           "cursor": "pointer",
           // in our test cases we use reason for the unique key
-          "data-testid": row.original.reason,
+          "data-testid": "req_",
```
JoshuaWilkes: is this intentional? it no longer matches the comment

web/src/components/tables/UsersTable.tsx
```diff
@@ -92,6 +92,13 @@ export const UsersTable = () => {
         emptyText: "No users",
         apiPaginator: paginator,
         linkTo: true,
+        rowProps: (row) => ({
+          "_hover": { bg: "gray.50" },
+          "cursor": "pointer",
+          // in our test cases we use reason for the unique key
```
JoshuaWilkes: comment doesn't match
Do we actually need all the row props like _hover ect to be specified? I though it would have been applied already 
```suggestion
          // in our test cases we use reason for the unique key
```

JoshuaWilkes: In my local branch the hover and pointer are already defaults

web/tests/approvals/approval.test.ts
```diff
@@ -60,40 +60,40 @@ test.describe.serial("Approval/Request Workflows", () => {
     await expect(locator).toContainText(uniqueReason);
   });
 
-  test("test approval workflow", async ({ page }) => {
-    // This will log us in as an admin
-    await LoginAdmin(page);
+  // test("test approval workflow", async ({ page }) => {
```
JoshuaWilkes: Need to uncomment this before merge

JoshuaWilkes (changes requested):

JoshuaWilkes (approved):

JoshuaWilkes:
PR Fixes a race condition in the auth context where initiating logout would sometimes log a user back in before it was complete.

The logoutURL in cognito now contains extra state so the frontend knows not to attempt authentication again

PR #375, opened 2022-11-22
Jordi/cf 306 web naming updates

jordiup:
related to renaming on #374


Eddie023 (approved):

JoshuaWilkes:
Implements a codebase wide rebrand from `granted approvals` to `common fate`

This includes some cleanup
- environment values for the application now all prefixed with COMMONFATE
- deployment config default changed from `granted-deployment.yml` to  `deployment.yml` 
- default administrators group changed to `common_fate_administrators` backwards compatibility is preserved in the infrastructure by keeping the `granted_administrators` group, this means we can phase it out at a later date, but all new deployments will use the new one.

For production deployments, a warning will be issues when running gdeploy informing about the name change to the config file.

For dev, this warning is shown as well. users will be required to change the name of the deployment config file manually before updating their stack.

PR #378, opened 2022-11-24
Revoke Grant Error Handling

jordiup:


JoshuaWilkes (commented):
DE = we create a new event type for revokeFailed rather than using GrantFailed

That way, we know the difference between a user action to revoke and a background workflow error.

I think the way we should implement this is to add an Error type into the framework that is returned from the revoke method.

This error type shoudl indicate that an error means the grant couldn;t be revoked normally, but that access is no longer active.

If the error is something else, then we shouldn;t cancel the workflow.

JoshuaWilkes:
I propose we park this for now, and revisit after some of our upcoming planning is done

PR #383, opened 2022-11-25
update asap grants to include request information

meyerjrr:
Screen Shot 2022-11-25 at 3 24 22 pm


JoshuaWilkes:
- It would be nice if this showed the expiry time rather thanASAP and the duration 1m
- We don't need to show the status of the request as part of this because it is already part of the message to the user

JoshuaWilkes (changes requested):
For this feature, I think you should create a new function that formats a message block specific to a user.

Also, this message needs to be sent as a single request to slack, currently it looks like you are sending a message then immediately updating it. which is an UDE

pkg/notifiers/slack/update.go
```diff
@@ -43,3 +44,26 @@ func (n *SlackNotifier) UpdateMessageBlockForReviewer(ctx context.Context, revie
 	}
 	return nil
 }
+
+// updates a message given a provided user id
+func (n *SlackNotifier) UpdateMessageBlockForRequester(ctx context.Context, requester *identity.User, message slack.Message) error {
```
JoshuaWilkes: Can this be removed now?

JoshuaWilkes (changes requested):
Could you please add some new screenshots with the latest messages that are displayed for both an auto approved and a request requiring approval?

meyerjrr:
Screen Shot 2022-11-28 at 3 57 01 pm
How it looks for scheduled requests and automatic requests

meyerjrr:
Also made a quick fix to the error boundary back button

JoshuaWilkes:
@meyerjrr is it possible to use this formatting for the expires at time? I think it should be more human readable. 

image

pkg/notifiers/slack/handle_request.go
```diff
@@ -149,18 +149,21 @@ func (n *SlackNotifier) HandleRequestEvent(ctx context.Context, log *zap.Sugared
 				wg.Wait()
 			}
 		} else {
+
 			//Review not required
-			msg := fmt.Sprintf(":white_check_mark: Your request to access *%s* has been automatically approved.", requestedRule.Name)
+			msg := fmt.Sprintf(":white_check_mark: Your request to access *%s* in account has been automatically approved.\n", requestedRule.Name)
```
JoshuaWilkes: I think we can omit the name of the rule here and instead add it to the table?
What do you think, given that the arguments are the most important information, saying that the request to <rule name> doesn't tell the user much

```suggestion
			msg := ":white_check_mark: Your request has been automatically approved.\n"
```

pkg/notifiers/slack/handle_request.go
```diff
@@ -149,18 +149,21 @@ func (n *SlackNotifier) HandleRequestEvent(ctx context.Context, log *zap.Sugared
...
 			if err != nil {
 				return errors.Wrap(err, "building review URL")
 			}
-			n.SendDMWithLogOnError(ctx, log, request.RequestedBy, msg, fallback)
+
+			n.sendRequestDetailsMessage(ctx, log, request, requestedRule, *requestingUserQuery.Result, msg, fallback)
 		}
 	case gevent.RequestApprovedType:
-		msg := fmt.Sprintf("Your request to access *%s* has been approved.", requestedRule.Name)
+		msg := fmt.Sprintf(":white_check_mark: Your request to access *%s* has been approved.", requestedRule.Name)
```
JoshuaWilkes, line 164: I think the same comment as above allies here on including rule name in the table instead

JoshuaWilkes (changes requested):
Proposing changes to the messaging

pkg/notifiers/slack/handle_request.go
meyerjrr: I don't see any issue with having it in the default text, we could additionally add it to the table?

JoshuaWilkes:
```suggestion
			msg := fmt.Sprintf(":white_check_mark: Your request to access *%s* has been automatically approved.\n", requestedRule.Name)
```

pkg/notifiers/slack/build_message.go
```diff
@@ -111,7 +111,7 @@ func BuildRequestMessage(o RequestMessageOpts) (summary string, msg slack.Messag
 
 	if o.WasReviewed || o.Request.Status == access.CANCELLED {
 		t := time.Now()
```
JoshuaWilkes, line 113: This already existed, shouldn't this be the time that the request was review, you can get this from the event input.
Rather than the time the event was processed (now())

JoshuaWilkes (approved):

PR #385, opened 2022-11-28
allow old stack name prefix

JoshuaWilkes:
Allows stackname to be prefixed with granted-approvals or common-fate in dev deployments for backwards compatibility


jordiup (approved):
LGTM 👍

PR #387, opened 2022-11-28
CF-315 User without correct access should not view Access Request Form

Eddie023:
### Summary
1. Throw forbidden error if the rule is not authorized for provided user. Don't render AccessRequestForm if the backend API response status is `forbidden` 

### Screenshots
Screen Shot 2022-11-28 at 2 10 34 PM

### Updated to 
Screen Shot 2022-11-29 at 12 37 58 PM


JoshuaWilkes (changes requested):
1. Looking good, I think we can make the message less direct, 
"Oops, either this rule doesn't exist, or you can't request access to it"

2. This implementation needs to be able to catch the case when someone is a reviewer for a rule, but they don't have access to request it. The API currently returns the rule with a 200, but if the user tried to request it, they would get a 401 error.

So I would look at adding a value in the API type for access rule, 

canRequest : boolean 

if it is false, then the user should see the same error page.

pkg/service/rulesvc/get.go
```diff
@@ -8,7 +8,7 @@ import (
 	"github.com/common-fate/common-fate/pkg/storage"
 )
 
-func (s *Service) GetRule(ctx context.Context, ID string, user *identity.User, isAdmin bool) (*rule.AccessRule, error) {
+func (s *Service) GetRule(ctx context.Context, ID string, user *identity.User, isAdmin bool) (*rule.AccessRuleWithCanRequest, error) {
```
JoshuaWilkes: Functionality looks good, can you please rename this struct so it's the same as some of our other services for consistency
```suggestion
func (s *Service) GetRule(ctx context.Context, ID string, user *identity.User, isAdmin bool) (*rule.GetAccessRuleResponse, error) {
```

JoshuaWilkes (changes requested):
Looks good, I would just like the Struct renamed for consistency

JoshuaWilkes (approved):

PR #397, opened 2022-12-16
CF-339: overflow fix

jordiup:
Fix to add horizontal overflow scroll on access rules container
https://linear.app/common-fate/issue/CF-339/fix-horizontal-scroll-on-ui


chrnorm (approved):

chrnorm:
Looks good, merged but I haven't verified this separately so we can quickly check it in our next UAT deployment

PR #399, opened 2022-12-20
Jordi/cf 338 keyboard friendly search

jordiup:


meyerjrr (commented):
Looks and works good so far @jordiup. Some questions from me:
Can we make the row items clickable by mouse too? Usually that is my first initial UX intuition after searching in a search box. 
Also I remember there was a "Switch to admin" button which I was a fan of in one of the previous looms, Has that been removed?

meyerjrr:
Also Node ci tests + playwright tests are failing fyi

jordiup:
@meyerjrr, tests are now passing. Have fixed the clickable issue. Think this one is g2g 🚀

JoshuaWilkes:
On long rule names, there is a scroll on the search box.
This would probably be better if it wrapped the text instead

image

web/src/components/nav/Navbar.tsx
```diff
@@ -70,6 +83,27 @@ export const Navbar: React.FC = () => {
...
+                // w="113px"
+                // px={2}
+                textAlign="left"
+                color="neutrals.600"
+                onClick={modal.onOpen}
+              >
+                <SearchIcon color="neutrals.600" boxSize="15px" ml={3} mr={2} />
+                Search
+                <Flex ml={2} mr={3} display="inline-flex">
+                  <Kbd>⌘</Kbd>
```
JoshuaWilkes: We need to support users of windows and linux, at the moment this only supports mac users who have the command key.

The chakra docs use something like this, however navigator.platform is deprecated, so I would recommend using a different approach, there are a few examples online for this.
https://github.com/chakra-ui/chakra-ui-docs/blob/main/src/components/algolia-search.tsx
```typescript
const ACTION_KEY_DEFAULT = ['Ctrl', 'Control']
const ACTION_KEY_APPLE = ['⌘', 'Command']

const [actionKey, setActionKey] = React.useState<string[]>(ACTION_KEY_APPLE)
  React.useEffect(() => {
    if (typeof navigator === 'undefined') return
    const isMac = /(Mac|iPhone|iPod|iPad)/i.test(navigator.platform)
    if (!isMac) {
      setActionKey(ACTION_KEY_DEFAULT)
    }
  }, [])
```



```suggestion
                  <Kbd>⌘</Kbd>
```

web/src/components/CommandPalette.tsx
```diff
@@ -0,0 +1,420 @@
...
+
+type ICommand = {
+  name: string;
+  icon: ComponentWithAs<"svg", IconProps>;
+  action: () => void;
+  isAdminOnly?: boolean;
+  type?: "favorite";
+};
+
+const EditActionModal = ({
```
JoshuaWilkes: I suspect this was copied over from power editor, I think we need to rename this for the context here
```suggestion
const CommandPallet = ({
```

JoshuaWilkes (changes requested):
A few changes to make but overall looking good

jordiup:
Cheers for the review Josh, think this one should now be G2G 👌 

image
image

JoshuaWilkes:
> Cheers for the review Josh, think this one should now be G2G 👌
> 
> image image

In the second image, the styling of the highlight looks to be breaking the layout, see the indent, I assume this is because its rendering as 2 elements side by side when highlighted?

jordiup:
> In the second image, the styling of the highlight looks to be breaking the layout, see the indent, I assume this is because its rendering as 2 elements side by side when highlighted?

Nice pickup Josh, fixed ✅ 
image
image

JoshuaWilkes (approved):

PR #400, opened 2022-12-20
CF-348: UI Artifact

jordiup:
Fix that takes Empty State outside of Grid wrapping component, should fix overflow scroll issue
https://linear.app/common-fate/issue/CF-348/ui-artefact-when-there-are-no-rule-assigned-to-a-user

### DIFF
--- a/web/src/pages/requests/index.tsx
+++ b/web/src/pages/requests/index.tsx
@@ -117,122 +117,119 @@ const Home = () => {
                       View All
                     </Button>
                   </Flex>
-                  <Grid
-                    mt={4}
-                    templateColumns={{
-                      base: "repeat(20, 1fr)",
-                      lg: "repeat(1, 1fr)",
-                      xl: "repeat(2, 1fr)",
-                    }}
-                    templateRows={{ base: "repeat(1, 1fr)", xl: "unset" }}
-                    minW={{ base: "unset", xl: "488px" }}
-                    gap={6}
-                    overflowX={{ base: "scroll" }}
-                  >
-                    {rules ? (
-                      rules.accessRules.length > 0 ? (
-                        rules.accessRules.map((r, i) => (
-                          <Link
-                            style={{ display: "flex" }}
-                            to={"/access/request/" + r.id}
-                            key={r.id}
-                          >
-                            <Box
-                              className="group"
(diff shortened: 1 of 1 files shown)


jordiup:
@meyerjrr small PR here. Just re-ran the tests and it's all looking good. Feel free to merge

web/src/pages/requests/index.tsx
```diff
@@ -117,122 +117,119 @@ const Home = () => {
...
-                            <Box
-                              className="group"
-                              textAlign="center"
-                              bg="neutrals.100"
-                              p={6}
-                              h="172px"
-                              w="232px"
-                              rounded="md"
-                              data-testid={"r_" + i}
+                  {((rules && rules.accessRules.length > 0) ||
```
JoshuaWilkes: @jordiup Would you be able to split this section out into its own component. Then, write the checks in a more declarative way so that its easier to read.

You may wish to break the component down some more and do the if statements differently.
The goal is to avoid complex logic inline for rendering, and prefer splitting into a seperate component for large sections of code

```
if (rules === undefined) {
return loading state
}
if (rules.length === 0){
return empty state
}
return standard state
```

JoshuaWilkes (changes requested):
Requesting that you split this code out into some components to improve readability and maintainability

web/src/pages/requests/index.tsx
jordiup: Cheers for the review and feedback @JoshuaWilkes. 

I see the benefits you've outlined regarding declaration and simplifying complex logic,

It's certainly a messy piece of logic, however I'm not convinced it can be simplified much more?

The loading/standard states are wrapped in a containing `<Grid>` component (which prevents re-rendering/code re-use), and then another set of conditional checks for loading/loaded occur within this. 

I've implemented the changes as per your suggestion, stripping it from inline and putting it into a component. 
https://github.com/common-fate/common-fate/commit/cab37581bfacc720a2782f2a2394eedad43c9e9c

Lmk what you think/feel free to merge 👌

JoshuaWilkes:
@jordiup FYI there is a build error on the FE

JoshuaWilkes (commented):
Happy with the changes, just need the build fixed

PR #408, opened 2023-01-10
Governance api

meyerjrr:
## Describe your changes

This PR adds an external developer/governance API. The purpose for creating this was to be able to interact with Common Fate via Terraform.
- Currently this adds functionality to be able to run CRUD operations for an access rule. This can be expanded on in the future for other aspects of the Common Fate product. 
-

## Type of change

## Issue and Documentation

- Documentation can be found on the readme of the terraform repo, and in internal Notion documents

## Testing

1.


web/src/components/tables/AccessRuleTable.tsx
```diff
@@ -132,7 +132,11 @@ export const AccessRuleTable = () => {
             <HStack>
               <UserAvatarDetails
                 tooltip
-                user={cell.value?.createdBy}
+                user={
+                  cell.value?.createdBy !== ""
+                    ? cell.value?.createdBy
+                    : "Common Fate Terraform Bot"
```
meyerjrr: Wasn't a massive fan of doing this, it would otherwise show up blank in the frontend. Keen to hear some feedback on this

chrnorm: @meyerjrr the "created by" field in the backend should be attributed to the Common Fate Terraform Bot, rather than just stored in the frontend IMO

meyerjrr: It is, but normally this is a calculated field (requires a user lookup via userId), then the users email is displayed in the frontend. But saving "Terraform Bot" to the `created_by` field causes this calculated field to break - resulting in a blank value. 

I Should have added this context in my initial comment.

pkg/api/api.go
```diff
@@ -100,13 +100,14 @@ type AccessService interface {
...
-	ArchiveAccessRule(ctx context.Context, user *identity.User, in rule.AccessRule) (*rule.AccessRule, error)
-	CreateAccessRule(ctx context.Context, user *identity.User, in types.CreateAccessRuleRequest) (*rule.AccessRule, error)
 	GetRule(ctx context.Context, ID string, user *identity.User, isAdmin bool) (*rule.GetAccessRuleResponse, error)
 	UpdateRule(ctx context.Context, in *rulesvc.UpdateOpts) (*rule.AccessRule, error)
 	RequestArguments(ctx context.Context, accessRuleTarget rule.Target) (map[string]types.RequestArgument, error)
 }
+
 type CacheService interface {
 	RefreshCachedProviderArgOptions(ctx context.Context, providerId string, argId string) (bool, []cache.ProviderOption, []cache.ProviderArgGroupOption, error)
 	LoadCachedProviderArgOptions(ctx context.Context, providerId string, argId string) (bool, []cache.ProviderOption, []cache.ProviderArgGroupOption, error)
```
chrnorm, line 113: Nitpick here (not sure if this is related to this specific PR) - what's the difference between `RefreshCachedProviderArgOptions` and `LoadCachedProviderArgOptions`? In the `validateTargetArgumentAgainstCachedOptions` method in this PR, we've swapped from one method to another, but I find it difficult to understand the difference between these two methods in a code review.

My assumption is that `LoadCachedProviderArgOptions` simply returns cached values, whereas `RefreshCachedProviderArgOptions` invalidates the cache.

Could we refactor these to be:

```go
type LoadArgOptionsOpts struct {
  ProviderID string
  ArgID string
  // RefreshCache invalidates the cache and refreshes argument options using the Access Provider.
  RefreshCache bool
}

type CacheService interface {
    LoadArgOptions(ctx context.Context, opts LoadArgOptionsOpts) (bool, []cache.ProviderOption, []cache.ProviderArgGroupOption, error)
}
```

chrnorm: My concern is that the existing implementation makes it very easy to mix up `RefreshCachedProviderArgOptions` with `LoadCachedProviderArgOptions`

web/src/components/UserAvatar.tsx
```diff
@@ -45,15 +46,29 @@ export const UserAvatarDetails: React.FC<UserAvatarProps> = ({
   tooltip,
   ...rest
 }) => {
+  const Component = tooltip ? TooltipAvatar : Avatar;
+
+  //skip lookup user step if creating user was terraform
+  if (user === "bot_governance_api") {
```
chrnorm: See the handling here for the `bot_governance_api` user

web/src/components/tables/AccessRuleTable.tsx
chrnorm: @meyerjrr isn't there already handling in the `UserAvatarDetails` component for this case?

My concern here by the way is that in future we might add additional types of service accounts - for example, a rule could be created by `bot_python_sdk`. This code here could cause bugs in future where "Common Fate Terraform Bot" is incorrectly rendered

web/src/components/UserAvatar.tsx
```diff
@@ -45,15 +46,29 @@ export const UserAvatarDetails: React.FC<UserAvatarProps> = ({
...
+          icon={<TerraformIcon fontSize="0.9rem" />}
+          {...rest}
+          bg="white"
+        />
+        <Text {...textProps}>Terraform</Text>
+      </HStack>
+    );
+  }
   // @ts-ignore
   const { data, isValidating } = useGetUser(user);
```
chrnorm: Could you please refactor this component so that the `useGetUser` hook is not conditionally called? (sorry, I'm pretty sure I initially wrote this component too!). @jordiup can give you a hand if you need.

The problem with this component is that if the `user` prop is equal to "bot_governance_api", the component returns early and the `useGetUser` hook is never called. This breaks one of the [Rules of Hooks](https://reactjs.org/docs/hooks-rules.html): "always use Hooks at the top level of your React function, before any early returns".

We'll need to split this component up into a `UserAvatar` and a `BotAvatar` component I think, and have a top level `CFAvatar` component that checks the `name` prop and renders one of the two

chrnorm (changes requested):
Looks great upon a full review, some minor nitpicks only

pkg/api/api.go
meyerjrr: I see where you are coming from here @chrnorm. I can make the change, but I think this will need to be refactored anyway with the inclusion of the new v2 providers provider arg options, this is what I rewrote today actually. 

I suggest we make a ticket to remind to update this when that is all merged in. (I'd say this will be a week or so)

chrnorm: Sounds good - will add this to our backlog

chrnorm (approved):

PR #410, opened 2023-01-12
Fix aws sso user check

meyerjrr:
## Describe your changes

- Previously when validating that a user existed in AWS SSO we were assuming the username field would always be an email. It was brought to our attention that this is not the case. This PR adjusts the logic to check against the Email field rather than username.

## Type of change

## Issue and Documentation

-

## Testing

1.

- [x] New and existing unit tests pass with my changes.

### DIFF
--- a/accesshandler/pkg/providers/aws/sso-v2/validate.go
+++ b/accesshandler/pkg/providers/aws/sso-v2/validate.go
@@ -43,7 +43,7 @@ func (p *Provider) ValidateGrant() providers.GrantValidationSteps {
 				}
 
 				if len(res.Users) != 0 {
-					return diagnostics.Info("User exists in SSO")
+					return diagnostics.Info("User exists in AWS SSO")
 				}
 
 				//Fallback attempt at finding a users email
--- a/accesshandler/pkg/providers/aws/sso-v2/validate.go
+++ b/accesshandler/pkg/providers/aws/sso-v2/validate.go
@@ -28,25 +28,34 @@ func (p *Provider) ValidateGrant() providers.GrantValidationSteps {
 				if err != nil {
 					return diagnostics.Error(err)
 				}
-				res, err := p.idStoreClient.ListUsers(ctx, &identitystore.ListUsersInput{
-					IdentityStoreId: aws.String(p.identityStoreID.Get()),
-					Filters: []types.Filter{{
-						AttributePath:  aws.String("UserName"),
-						AttributeValue: aws.String(subject),
-					}},
-				})
-				if err != nil {
-					return diagnostics.Error(err)
+
+				//Pull all users and check if emails match to find if user exists in AWS SSO
+				//This was required as filtering on Username, does not always work since some users do not use email as their username in AWS SSO
+				hasMore := true
+				var nextToken *string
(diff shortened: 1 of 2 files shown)


accesshandler/pkg/providers/aws/sso-v2/validate.go
```diff
@@ -28,25 +28,37 @@ func (p *Provider) ValidateGrant() providers.GrantValidationSteps {
...
-					return diagnostics.Error(err)
-				}
 
-				if len(res.Users) == 0 {
-					return diagnostics.Error(fmt.Errorf("could not find user %s in AWS SSO", subject))
-				}
-				if len(res.Users) > 1 {
-					// this should never happen, but check it anyway.
-					return diagnostics.Error(fmt.Errorf("expected 1 user but found %v", len(res.Users)))
+				//Pull all users and check if emails match to find if user exists in AWS SSO
```
chrnorm, line 50: @meyerjrr even if this check passes, doesn't our actual Grant method still use the user's username, rather than their email address? So validation would pass but the Grant method would still fail?

You should be able to test this in our sandbox org by creating a user with a prefix username such as `jack.meyer` and seeing whether granting will still work.

Additionally, calling `ListUsers` is going to make grant validation a lot slower for everyone (in some cases it could take 5-10s+ to iterate through all users for higher user counts). At the least it might be good to make this option a config flag on the provider. Ideally we would store a map of user email addresses to AWS SSO user IDs to make this lookup faster.

chrnorm (changes requested):

accesshandler/pkg/providers/aws/sso-v2/validate.go
meyerjrr: Ah yep, good call. The granting logic would need to change as well as this is retrieving a user based on the username being an email.

I'm thinking that this shouldn't really need to be a config option (unless its possible for users to not have any emails on their user objects in AWS SSO) and in that case we would need to update how we relate IDP users to AWS SSO users.

I'm thinking we can look up users filtered on the username first (for usernames which are emails) then fallback on a manual lookup through a list of SSO users. 
- +1 to the map of users idea to speed things up.

chrnorm: Sounds good to the username lookup first, then falling back to listing SSO users. In our next Access Provider iteration for Community Providers we can include the user mapping which should make this a lot faster too

accesshandler/pkg/providers/aws/sso-v2/validate.go
```diff
@@ -39,14 +42,42 @@ func (p *Provider) ValidateGrant() providers.GrantValidationSteps {
 					return diagnostics.Error(err)
 				}
 
-				if len(res.Users) == 0 {
-					return diagnostics.Error(fmt.Errorf("could not find user %s in AWS SSO", subject))
+				if len(res.Users) != 0 {
+					return diagnostics.Info("User exists in SSO")
```
chrnorm:
```suggestion
					return diagnostics.Info("User exists in AWS SSO")
```

chrnorm (approved):

PR #411, opened 2023-01-12
prevent a nil panic error for some cases in identity syncing

chrnorm:
## Describe your changes

- A Common Fate user has reported a runtime error in the identity sync Lambda function, due to a nil pointer assignment to the `internalGroupUsers` variable. This PR fixes the error.

## Type of change

## Issue and Documentation

-

## Testing

1.

### DIFF
--- a/pkg/identity/identitysync/sync.go
+++ b/pkg/identity/identitysync/sync.go
@@ -225,6 +225,12 @@ func processUsersAndGroups(idpType string, idpUsers []identity.IDPUser, idpGroup
 			gid := ddbGroupMap[idpGroupId].ID
 			internalGroupIds[gid] = gid
 			uid := ddbUserMap[idpUser.Email].ID
+
+			// prevents a panic due to an assignment to a nil map
+			if internalGroupUsers[gid] == nil {
+				internalGroupUsers[gid] = map[string]string{}
+			}
+
 			internalGroupUsers[gid][uid] = uid
 		}
 
--- a/pkg/identity/identitysync/sync.go
+++ b/pkg/identity/identitysync/sync.go
@@ -221,8 +221,7 @@ func processUsersAndGroups(idpType string, idpUsers []identity.IDPUser, idpGroup
 
 		// This map ensures we have a distinct list of ids
 		internalGroupIds := map[string]string{}
-		for _, idpGroupId := range idpUser.Groups {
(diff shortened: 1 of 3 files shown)


pkg/identity/identitysync/sync.go
```diff
@@ -225,6 +225,12 @@ func processUsersAndGroups(idpType string, idpUsers []identity.IDPUser, idpGroup
 			gid := ddbGroupMap[idpGroupId].ID
 			internalGroupIds[gid] = gid
 			uid := ddbUserMap[idpUser.Email].ID
+
+			// prevents a panic due to an assignment to a nil map
+			if internalGroupUsers[gid] == nil {
```
JoshuaWilkes, line 229: @chrnorm this error would imply to me that there are group ids on the user object returned from the identity provider that are not returned by the groups api for the identity provider?

Is it possible that a user in google can have groups that are not returned by the groups API?

JoshuaWilkes: This code is all a little more complicated than it needs to be, A while back we used a different internal id to teh idp id for groups. 

Now we just use the external ID for both. In this sync function we seen to interchange between internal ID and external ID even though they should be the same.

This makes it harder to read :( but also shouldn't contribute to this error happening so i'm at a loss for why it happened

chrnorm: @JoshuaWilkes I have been trying to further debug this today and have also found it quite difficult to interpret the group syncing logic.

I have added an additional 100 groups to our internal Google Workspace test environment to see whether it could be a pagination issue but syncing works fine and I can't replicate the issue that our community member has. The pagination handling looked fine to me too during a review of the code today.

I have slightly simplified the group membership sync logic and added a unit test around the panic in [this commit](https://github.com/common-fate/common-fate/pull/411/commits/c5115b13403d2b393db450ce205b611e50426b3c). 

At the least, this code is now more defensive - if for whatever reason a group doesn't exist, the entire sync procedure will no longer fail. Rather, a member will not belong to the non-existent group and the sync will succeed. I think this behaviour is a lot more desirable as Common Fate remains partially usable, rather than entirely unusable in this scenario.

I propose that we merge this improvement and continue to look further into this if the user still has issues. What do you think?

Also, a quick look over my changes to the unit tests would be much appreciated.

JoshuaWilkes (approved):
Happy to go with this

PR #431, opened 2023-02-01
CF-602 Implement resource fetching

Eddie023:
## Describe your changes
- If the provider schema's audit Object has the "loadResources" nested object then for each task there we should recursively call them unless we get the resources. 
- Store all the fetched resources into dynamodb 

## Type of change
- [ New feature (non-breaking change which adds functionality)


pkg/cache/provider_resource.go
```diff
@@ -0,0 +1,24 @@
+package cache
+
+import (
+	"github.com/common-fate/common-fate/pkg/storage/keys"
+	"github.com/common-fate/ddb"
+)
+
+const PK_PREFIX = "PROVIDER_RESOURCE#"
```
JoshuaWilkes:
```suggestion
```

JoshuaWilkes (approved):

PR #432, opened 2023-02-02
add debug logging for Okta IDP sync

chrnorm:
Adds additional debugging logic so that 

```
CF_LOG=debug gdeploy identity sso enable
```

can be run to help identify issues during IDP sync.

### DIFF
--- a/pkg/deploy/config.go
+++ b/pkg/deploy/config.go
@@ -22,7 +22,6 @@ import (
 	"github.com/pkg/errors"
 	"github.com/segmentio/ksuid"
 	"github.com/urfave/cli/v2"
-	"go.uber.org/zap"
 	"gopkg.in/yaml.v3"
 )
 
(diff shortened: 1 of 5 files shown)


pkg/identity/identitysync/okta.go
```diff
@@ -50,22 +55,50 @@ func (s *OktaSync) TestConfig(ctx context.Context) error {
...
+	userJSON, err := json.Marshal(oktaUser)
+	if err != nil {
+		log.Errorw("error marshalling user for logging", zap.Error(err))
+	}
+
+	log.Debugw("converting okta user to internal user", "oktaUser", string(userJSON))
+
+	firstName, ok := (*oktaUser.Profile)["firstName"].(string)
+	if !ok {
+		log.Error("okta profile had no firstName")
```
JoshuaWilkes, line 74: If these are intended to be non blocking errors, we should add a comment in the code so we know why

pkg/deploy/config.go
```diff
@@ -245,16 +244,8 @@ func UnmarshalProviderMap(data string) (ProviderMap, error) {
 // RunConfigTest runs ConfigTest() if it is implemented on the interface
 func RunConfigTest(ctx context.Context, testable interface{}) error {
 	if tester, ok := testable.(gconfig.Tester); ok {
-		si := spinner.New(spinner.CharSets[14], 100*time.Millisecond)
```
JoshuaWilkes, line 248: Does CLIO automatically format the global logger now so these look nice even when using zap.S().Info("message")

JoshuaWilkes (commented):
A few comments on logging and error handling

pkg/deploy/config.go
chrnorm: Not automatically, but I used 

```
zap.ReplaceGlobals(clio.G())
```

in the entrypoint for `gdeploy` so that calls to `zap.S()` are formatted by the `clio` formatter

pkg/identity/identitysync/okta.go
chrnorm: Thanks, I've added a comment here

jordiup (approved):
Looks good to me 👍

PR #437, opened 2023-02-07
API Layer, Service Layer, Database Layer for targetGroup and Deployments

meyerjrr:
## Describe your changes

-

## Type of change

## Issue and Documentation

-

## Testing

1.


openapi.yml
```diff
@@ -1131,6 +1131,101 @@ paths:
       description: Update a favorite with new FavoriteDetails
       requestBody:
         $ref: "#/components/requestBodies/CreateFavoriteRequest"
+  "/api/v1/target-group-deployments/{id}":
+    get:
+      summary: Your GET endpoint
```
JoshuaWilkes: Prior to merging, we need to update the descriptions on the APIs and Models

The ids of the routes look good, nice work

```suggestion
      summary: Your GET endpoint
```

openapi.yml
```diff
@@ -2062,6 +2214,150 @@ components:
...
+          type: boolean
+        diagnostics:
+          type: array
+          items:
+            $ref: "#/components/schemas/TargetGroupDiagnostic"
+        activeConfig:
+          type: object
+          additionalProperties:
+            $ref: "#/components/schemas/TargetGroupDeploymentConfig"
+        provider:
```
JoshuaWilkes, line 2246: You may need to define this as its own model, otherwise it generates an an anonymous struct and it annoying to instantiate

openapi.yml
```diff
@@ -2497,6 +2824,43 @@ components:
...
+            type: object
+            properties:
+              targetSchema:
+                type: string
+              ID:
+                type: string
+            required:
+              - targetSchema
+              - ID
+    CreateTargetGroupDeploymentRequest:
```
JoshuaWilkes, line 2848: create deployment should also have an ID field which is specified by the user

pkg/api/targetgroup.go
```diff
@@ -0,0 +1,84 @@
...
+	var createGroupRequest types.CreateTargetGroupRequest
+	err := apio.DecodeJSONBody(w, r, &createGroupRequest)
+	if err != nil {
+		apio.Error(ctx, w, apio.NewRequestError(err, http.StatusBadRequest))
+		return
+	}
+	group, err := a.TargetGroupService.CreateTargetGroup(ctx, createGroupRequest)
+	if err == targetgroupsvc.ErrTargetGroupIdAlreadyExists {
+		// the user supplied id already exists
+		err = apio.NewRequestError(err, http.StatusBadRequest)
```
JoshuaWilkes: I would prefer that this is implemented like so

```suggestion
		apio.Error(ctx, w, apio.NewRequestError(err, http.StatusBadRequest))
		return
```

pkg/api/targetgroup.go
```diff
@@ -0,0 +1,84 @@
...
+// (GET /api/v1/target-groups/{id})
+func (a *API) GetTargetGroup(w http.ResponseWriter, r *http.Request, id string) {
+	ctx := r.Context()
+
+	q := storage.GetTargetGroup{ID: id}
+
+	_, err := a.DB.Query(ctx, &q)
+
+	if err == ddb.ErrNoItems {
+		err = apio.NewRequestError(err, http.StatusNotFound)
```
JoshuaWilkes: I would prefer that this is implemented like so
```suggestion
		apio.Error(ctx, w, apio.NewRequestError(err, http.StatusNotFound))
		return
```

pkg/api/targetgroup_deployment.go
```diff
@@ -0,0 +1,122 @@
...
+
+	ctx := r.Context()
+	res := types.ListTargetGroupDeploymentsResponse{
+		Res: []types.TargetGroupDeployment{},
+	}
+
+	listTargetGroupDeployments := storage.ListTargetGroupDeployments{}
+
+	dbq, err := a.DB.Query(ctx, &listTargetGroupDeployments)
+	if err != nil {
```
JoshuaWilkes: If this return ddb.ErrNoItems then the API will respond in 500

Instead check if the error is !ddb.ErrNoItems
```
	if err != nil && err != ddb.ErrNoItems {
		apio.Error(ctx, w, err)
		return
	}
```

pkg/api/targetgroup_deployment.go
```diff
@@ -0,0 +1,122 @@
...
+	// TODO: run pre-lim checks to ensure aws account/arn are valid
+
+	dbInput := targetgroup.Deployment{
+		ID:           types.NewTargetGroupDeploymentID(),
+		FunctionARN:  b.FunctionArn,
+		Runtime:      b.Runtime,
+		AWSAccount:   b.AwsAccount,
+		Healthy:      false,
+		Diagnostics:  []targetgroup.Diagnostic{},
+		ActiveConfig: map[string]targetgroup.Config{},
```
JoshuaWilkes: The active config will be set by a healthcheck service which runs async. when creating a new deployment, it can be left empty
```suggestion
```

pkg/api/targetgroup_deployment.go
```diff
@@ -0,0 +1,122 @@
...
+		Provider:     targetgroup.Provider{},
```
JoshuaWilkes: The provider will be set by a healthcheck service which runs async. when creating a new deployment, it can be left empty
```suggestion
```

pkg/api/targetgroup_deployment.go
```diff
@@ -0,0 +1,122 @@
...
+
+	// TODO: run pre-lim checks to ensure aws account/arn are valid
+
+	dbInput := targetgroup.Deployment{
+		ID:           types.NewTargetGroupDeploymentID(),
+		FunctionARN:  b.FunctionArn,
+		Runtime:      b.Runtime,
+		AWSAccount:   b.AwsAccount,
+		Healthy:      false,
+		Diagnostics:  []targetgroup.Diagnostic{},
```
JoshuaWilkes: The default status for a new deployment is unhealthy. This is because the deployment has never been healthchecked.

A deployment which cannot be reached should have a Diagnostic log that says it is not online.

Define this diagnostic log in the targetgroup package and use it here

```suggestion
		Diagnostics:  []targetgroup.Diagnostic{},
```

pkg/api/targetgroup_deployment.go
```diff
@@ -0,0 +1,122 @@
...
+		Provider:     targetgroup.Provider{},
+	}
+
+	dbInput.Provider.Name = b.Provider.Name
+	dbInput.Provider.Version = b.Provider.Version
+	dbInput.Provider.Version = b.Provider.Version
+
+	/**
+
+	TODO
```
JoshuaWilkes: See my comments above, this is not needed

pkg/api/targetgroup_deployment.go
```diff
@@ -0,0 +1,122 @@
...
+	err := apio.DecodeJSONBody(w, r, &b)
+	if err != nil {
+		apio.Error(ctx, w, err)
+		return
+	}
+
+	// TODO: run pre-lim checks to ensure aws account/arn are valid
+
+	dbInput := targetgroup.Deployment{
+		ID:           types.NewTargetGroupDeploymentID(),
```
JoshuaWilkes: The ID should be user supplied via the API

```suggestion
		ID:           b.ID
```

pkg/api/targetgroup_deployment.go
```diff
@@ -0,0 +1,122 @@
...
+
+	// @TODO: run a check here to ensure no overwrites occur ...
+
+	err = a.DB.Put(ctx, &dbInput)
+	if err != nil {
+		apio.Error(ctx, w, err)
+		return
+	}
+
+	if err != nil {
```
JoshuaWilkes, line 65: Remove duplicate error block

pkg/api/targetgroup_deployment.go
```diff
@@ -0,0 +1,122 @@
...
+// Your GET endpoint
+// (GET /api/v1/target-group-deployments/{id})
+func (a *API) GetTargetGroupDeployment(w http.ResponseWriter, r *http.Request, id string) {
+
+	ctx := r.Context()
+
+	q := storage.GetTargetGroupDeployment{ID: id}
+
+	_, err := a.DB.Query(ctx, &q)
+	if err != nil {
```
JoshuaWilkes, line 88: Check for ddb.ErrNoItems and return a 404 error if the deployment is not found

```suggestion
	if err != nil {
```

pkg/service/targetdeploymentsvc/service.go
```diff
@@ -0,0 +1,20 @@
+package targetdeploymentsvc
+
+import (
+	"github.com/common-fate/ddb"
+	"k8s.io/utils/clock"
+)
+
+// Service holds business logic relating to Cognito user management.
```
JoshuaWilkes, line 8:
```suggestion
// Service holds business logic relating to Target Group Deployments
```

pkg/service/targetgroupsvc/create.go
```diff
@@ -0,0 +1,36 @@
...
+)
+
+func (s *Service) CreateTargetGroup(ctx context.Context, req types.CreateTargetGroupRequest) (*targetgroup.TargetGroup, error) {
+	log := zap.S()
+
+	//look up target schema for the provider version
+
+	splitKey := strings.Split(req.TargetSchema, "/")
+
+	resp, err := s.ProviderRegistryClient.GetProviderWithResponse(ctx, splitKey[0], splitKey[1], splitKey[2])
```
JoshuaWilkes, line 27: You will need to validate here that the length of the split target is what you expect

pkg/service/targetgroupsvc/service.go
```diff
@@ -0,0 +1,25 @@
...
+)
+
+type Service struct {
+	Clock                  clock.Clock
+	DB                     ddb.Storage
+	ProviderRegistryClient registry_types.ClientWithResponsesInterface
+}
+
+//go:generate go run github.com/golang/mock/mockgen -destination=mocks/targetgroup.go -package=mocks . TargetGroupService
+type TargetGroupService interface {
```
JoshuaWilkes: In Go and as our own best practice, we choose to add interfaces in the consumer or the interface not in the implementation.

What this means is that if this interface and the generated mocks are used by the API service, then this interface should be defined in the API package instead

pkg/service/targetgroupsvc/update.go
```diff
@@ -0,0 +1,58 @@
...
+
+func (s *Service) UpdateTargetGroup(ctx context.Context, req UpdateOpts) (*targetgroup.TargetGroup, error) {
+	log := zap.S()
+
+	//get target group for updating
+
+	q := storage.GetTargetGroup{ID: req.TargetGroupID}
+
+	_, err := s.DB.Query(ctx, &q)
+	if err != nil && err != ddb.ErrNoItems {
```
JoshuaWilkes: This ddb.ErrNoItems needs to be handled.

One way to handle this is to pass the targetGroup in as an input to the update method.

You would fetch the targetGroup in the api layer and return a 404 error if its not found, and the service just validates the input and updates the record

pkg/service/targetgroupsvc/update.go
```diff
@@ -0,0 +1,58 @@
...
+		return nil, err
+	}
+
+	targetGroup := q.Result
+
+	targetGroup.ID = req.UpdateRequest.ID
```
JoshuaWilkes: ID cannot be changed, and this assignment would have no effect anyway

```suggestion
```

pkg/service/targetgroupsvc/update.go
```diff
@@ -0,0 +1,58 @@
...
+	"github.com/common-fate/ddb"
+	"go.uber.org/zap"
+)
+
+type UpdateOpts struct {
+	TargetGroupID string
+	UpdateRequest types.CreateTargetGroupRequest
+}
+
+func (s *Service) UpdateTargetGroup(ctx context.Context, req UpdateOpts) (*targetgroup.TargetGroup, error) {
```
JoshuaWilkes: Based on the linear ticket https://linear.app/common-fate/issue/CF-624/create-api-methods-for-targetgroup-and-deployment-as-per-spec

The update method is not a requirement for this PR, you should exclude it as the behaviour is not define in our designs

pkg/service/targetgroupsvc/update.go
JoshuaWilkes: The update method can be removed, see above comment

pkg/service/targetgroupsvc/update.go
JoshuaWilkes: The update method can be removed, see above comment

pkg/service/targetgroupsvc/update_test.go
```diff
@@ -0,0 +1,95 @@
...
+	"github.com/common-fate/common-fate/pkg/types"
+	"github.com/common-fate/ddb/ddbmock"
+	"github.com/common-fate/provider-registry-sdk-go/pkg/providerregistrysdk"
+	"github.com/common-fate/provider-registry-sdk-go/pkg/providerregistrysdk/prmocks"
+	"github.com/stretchr/testify/assert"
+
+	"github.com/golang/mock/gomock"
+)
+
+func TestUpdateTargetGroup(t *testing.T) {
```
JoshuaWilkes: Based on the linear ticket here https://linear.app/common-fate/issue/CF-624/create-api-methods-for-targetgroup-and-deployment-as-per-spec update target group is not a requirement for this PR as the behaviour is not defined in our designs

pkg/storage/get_target_deployment.go
```diff
@@ -0,0 +1,25 @@
...
+)
+
+type GetTargetGroupDeployment struct {
+	ID     string
+	Result targetgroup.Deployment `ddb:"result"`
+}
+
+func (g *GetTargetGroupDeployment) BuildQuery() (*dynamodb.QueryInput, error) {
+	qi := dynamodb.QueryInput{
+		KeyConditionExpression: aws.String("PK = :pk AND begins_with(SK, :sk)"),
```
JoshuaWilkes, line 18: This needs to have a limit 1 expression 
```
Limit:                  aws.Int32(1),
```
and should also include 
```
func (g *GetGroup) UnmarshalQueryOutput(out *dynamodb.QueryOutput) error {
	if len(out.Items) != 1 {
		return ddb.ErrNoItems
	}

	return attributevalue.UnmarshalMap(out.Items[0], &g.Result)
}
```

pkg/storage/get_target_group.go
```diff
@@ -0,0 +1,25 @@
...
+	"github.com/common-fate/common-fate/pkg/storage/keys"
+	"github.com/common-fate/common-fate/pkg/targetgroup"
+)
+
+type GetTargetGroup struct {
+	ID     string
+	Result targetgroup.TargetGroup `ddb:"result"`
+}
+
+func (g *GetTargetGroup) BuildQuery() (*dynamodb.QueryInput, error) {
```
JoshuaWilkes, line 18: This needs to have a limit 1 expression 
```
Limit:                  aws.Int32(1),
```
and should also include 
```
func (g *GetGroup) UnmarshalQueryOutput(out *dynamodb.QueryOutput) error {
	if len(out.Items) != 1 {
		return ddb.ErrNoItems
	}

	return attributevalue.UnmarshalMap(out.Items[0], &g.Result)
}
```

pkg/targetgroup/deployment.go
```diff
@@ -0,0 +1,76 @@
...
+		diagnostics[i] = types.TargetGroupDiagnostic{
+			Code:    d.Code,
+			Level:   d.Level,
+			Message: d.Message,
+		}
+	}
+
+	// new struct that can be cast to json version of Provdier
+	// Q: should this instead be declared in the types package?
+	provider := struct {
```
JoshuaWilkes: This should be defined in the type package as its own model, see my other comments

pkg/targetgroup/targetgroup.go
```diff
@@ -0,0 +1,101 @@
...
+type TargetGroup struct {
+	//user defined e.g. 'okta'
+	ID           string            `json:"id" dynamodbav:"id"`
+	TargetSchema GroupTargetSchema `json:"grouptargetSchema" dynamodbav:"groupTargetSchema"`
+	// reference to the SVG icon for the target group
+	Icon              string                   `json:"icon" dynamodbav:"icon"`
+	TargetDeployments []DeploymentRegistration `json:"targetDeployments" dynamodbav:"targetDeployments"`
+}
+
+type GroupTargetSchema struct {
```
JoshuaWilkes, line 23: Can this just be called TargetSchema?

pkg/targetgroup/targetgroup.go
```diff
@@ -0,0 +1,101 @@
...
+
+func (r *TargetGroup) DDBKeys() (ddb.Keys, error) {
+	keys := ddb.Keys{
+		PK: keys.TargetGroup.PK1,
+		SK: keys.TargetGroup.SK1(r.ID),
+	}
+	return keys, nil
+}
+
+func (r *GroupTargetSchema) ToAPI() types.GroupTargetSchema {
```
JoshuaWilkes: If this needs to be prefixed to avoid a naming collision, please name the API model TargetGroupTargetSchema

But leave the Struct name in this file as TargetSchema

```suggestion
func (r *GroupTargetSchema) ToAPI() types.TargetGroupTargetSchema {
```

pkg/types/resource_id.go
```diff
@@ -49,3 +49,7 @@ func NewProviderSetupID() string {
 func NewDeploymentID() string {
 	return newResourceID("dep")
 }
+
+func NewTargetGroupDeploymentID() string {
```
JoshuaWilkes: This is not required as IDs are user supplied
```suggestion

```

JoshuaWilkes (changes requested):

pkg/targetgroup/targetgroup.go
meyerjrr: This was decided as I found it painful referencing both `targetgroup.TargetSchema` and `providerregistrysdk.TargetSchema` in the same files.

openapi.yml
jordiup: Have just added some preliminary descriptions to each end point

openapi.yml
jordiup: I think I have now sussed this? Basically the openapi spec becomes a bit more complex, but you get the nicer instantiating in go 👌 

image

pkg/api/targetgroup_deployment.go
```diff
@@ -0,0 +1,83 @@
...
+	ctx := r.Context()
+
+	var b types.CreateTargetGroupDeploymentRequest
+	err := apio.DecodeJSONBody(w, r, &b)
+	if err != nil {
+		apio.Error(ctx, w, err)
+		return
+	}
+
+	req, err := a.TargetGroupDeploymentService.CreateTargetGroupDeployment(ctx, b)
```
JoshuaWilkes: Any validation errors returned from this should be handled as a 400 response not a 500

Tests should validate this

pkg/service/targetdeploymentsvc/create.go
```diff
@@ -0,0 +1,57 @@
...
+		FunctionARN:  req.FunctionArn,
+		Runtime:      req.Runtime,
+		AWSAccount:   req.AwsAccount,
+		Healthy:      false,
+		Diagnostics:  []targetgroup.Diagnostic{},
+		ActiveConfig: map[string]targetgroup.Config{},
+		Provider:     targetgroup.Provider{},
+	}
+
+	dbInput.Provider.Name = req.Provider.Name
```
JoshuaWilkes: This will come from the healthcheck service, so it should be removed here. It should also be removed from the API request

JoshuaWilkes (changes requested):

PR #444, opened 2023-02-08
Bootstrap provider

meyerjrr:
## Describe your changes

- Implements bootstrap provider command which copies the assets from the published assets bucket to the bootstrap bucket

## Type of change

## Issue and Documentation

-

## Testing

1.

### DIFF
--- a/cf/cmd/cli/commands/provider/provider.go
+++ b/cf/cmd/cli/commands/provider/provider.go
@@ -8,12 +8,12 @@ import (
 	"path"
 	"strings"
 
+	aws_config "github.com/aws/aws-sdk-go-v2/config"
+	"github.com/aws/aws-sdk-go-v2/service/s3"
+	"github.com/aws/aws-sdk-go/aws"
 	"github.com/common-fate/clio"
-	"github.com/common-fate/common-fate/cf/pkg/bootstrapper"
-	"github.com/common-fate/common-fate/pkg/config"
+	"github.com/common-fate/common-fate/internal/build"
 	"github.com/common-fate/provider-registry-sdk-go/pkg/providerregistrysdk"
-	"github.com/joho/godotenv"
-	"github.com/sethvargo/go-envconfig"
 	"github.com/urfave/cli/v2"
 )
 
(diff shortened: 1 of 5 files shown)


cf/cmd/cli/commands/provider/provider.go
```diff
@@ -19,11 +30,72 @@ var BootstrapCommand = cli.Command{
 	Name:        "bootstrap",
 	Description: "bootstrap a provider from the registry",
 	Usage:       "bootstrap a provider from the registry",
+	Flags: []cli.Flag{
+		&cli.StringFlag{Name: "bootstrapBucket", Aliases: []string{"bb"}, Usage: "The name of the bootstrap bucket provider is deployed to", EnvVars: []string{"DEPLOYMENT_BUCKET"}},
```
JoshuaWilkes:
```suggestion
		&cli.StringFlag{Name: "bootstrapBucket", Aliases: []string{"bb"}, Usage: "The name of the bootstrap bucket to copy assets into", EnvVars: []string{"DEPLOYMENT_BUCKET"}},
```

cf/cmd/cli/commands/provider/provider.go
```diff
@@ -19,11 +30,72 @@ var BootstrapCommand = cli.Command{
...
+		keys := strings.Split(id, "@")
+
+		if len(keys) != 2 {
+			return errors.New("incorrect provider id given")
+		}
+
+		version := keys[1]
+
+		teamAndName := strings.Split(keys[0], "/")
+
```
JoshuaWilkes, line 63: You should add another length check here.
I think we will want to add a parser that can parse a provider string and return an object containing the components

cf/cmd/cli/commands/provider/provider.go
```diff
@@ -19,11 +30,72 @@ var BootstrapCommand = cli.Command{
...
+		//get bootstrap bucket
+
+		//read from flag
+		bootstrapBucket := c.String("bootstrapBucket")
+
+		//work out the lambda asset path
+		lambdaAssetPath := path.Join(team, name, version)
+
+		//copy the provider assets into the bucket (this will also copy the cloudformation template too)
+		err = bs.CopyProviderAsset(ctx, res.JSON200.LambdaAssetS3Arn, lambdaAssetPath, bootstrapBucket)
```
JoshuaWilkes: Given that this function just runs an s3 copy, It will be acceptable to just put the s3 client and runction call directly in this command rather than putting it in the bootstrapper package

Done = shift copy into this file from bootstrapper

cf/cmd/cli/commands/provider/provider.go
```diff
@@ -19,11 +30,72 @@ var BootstrapCommand = cli.Command{
...
+	Flags: []cli.Flag{
+		&cli.StringFlag{Name: "bootstrapBucket", Aliases: []string{"bb"}, Usage: "The name of the bootstrap bucket provider is deployed to", EnvVars: []string{"DEPLOYMENT_BUCKET"}},
+	},
 	Action: func(c *cli.Context) error {
 		id := c.Args().First()
 		if id == "" {
 			return errors.New("id argument must be provided")
 		}
+
+		var cfg config.ProviderDeploymentCLI
```
JoshuaWilkes: Rather than loading config from a .env file to get the registry API URL

We will want to use internal/build/urls.go

```
package build

// these URLs are overridden as part of the release build process.
var (
	ProviderRegistryAPIURL = "localhost:9091"
)
```

at build time when we release this CLI, we will us `-ldflags` to overwrite the value of this variable with the production registry URL

To see this in action, we use this technique in the Granted CLI repo

JoshuaWilkes (changes requested):

JoshuaWilkes (approved):

PR #467, opened 2023-02-14
CF-658 implement targetgroup for accessrule

Eddie023:
## Describe your changes

1. Pass targetGroupId as provider Id where required. 
2. Add a condition to check if the provided ID is indeed a target group. If `targetgroup` then they require different handling.


pkg/service/rulesvc/create.go
```diff
@@ -143,11 +181,22 @@ func (s *Service) CreateAccessRule(ctx context.Context, userID string, in types.
...
-	target, err := s.ProcessTarget(ctx, in.Target)
-	if err != nil {
+	q := storage.GetTargetGroup{ID: in.Target.ProviderId}
+	_, err := s.DB.Query(ctx, &q)
+	if err != nil && err != ddb.ErrNoItems {
 		return nil, err
 	}
 
+	var target rule.Target
+	if q.Result.ID != "" {
```
JoshuaWilkes, line 191: Can you replace this if statement by just passing a boolean into the function
Seems like the target group query should return a pointer but its not,

pkg/service/rulesvc/update.go
```diff
@@ -17,9 +17,19 @@ type UpdateOpts struct {
 
 func (s *Service) UpdateRule(ctx context.Context, in *UpdateOpts) (*rule.AccessRule, error) {
 	clk := s.Clock
-	target, err := s.ProcessTarget(ctx, in.UpdateRequest.Target)
-	if err != nil {
-		return nil, err
+
+	var target rule.Target
+	var err error
+	if in.Rule.Target.TargetGroupID != "" {
```
JoshuaWilkes, line 22: Same as above, replace if statement with a variable assignment and don't duplicate the function call

JoshuaWilkes (changes requested):

JoshuaWilkes (approved):

PR #472, opened 2023-02-14
CF-605: fix menulist filtering issue

jordiup:
## Describe your changes

> The dropdown is not correctly indexed and shows ‘pending only’ when “all” is selected.

This should fix the issue described. The menu list wasn't getting the input value passed down to the Options/List child component. 

## Type of change

## Issue and Documentation

- https://linear.app/common-fate/issue/CF-605/access-requests-filtering-issue

## Testing

I've confirmed that the undesired behaviour has been fixed. I have however noticed additional strange behaviour. In certain states the value seems to hang on screen. This could be due to async logic in the navigate/router, or API query. I.e. 
```
        <RequestsFilterMenu
          onChange={(s) =>
            navigate({
              search: (old) => ({
                ...old,
                status: s?.toLowerCase() as Lowercase<RequestStatus>,
              }),
            })
          }
```

### DIFF
--- a/web/src/components/tables/RequestsFilterMenu.tsx
+++ b/web/src/components/tables/RequestsFilterMenu.tsx
@@ -37,6 +37,18 @@ export const RequestsFilterMenu: React.FC<{
           defaultValue="all"
           title="View option"
           type="radio"
+          value={
+            // map the status input to a value so that the MenuList hydrates the current state
+            status === "PENDING"
+              ? "pend"
+              : status === "DECLINED"
+              ? "den"
+              : status === "APPROVED"
+              ? "apr"
+              : status === "CANCELLED"
+              ? "can"
+              : "all"
+          }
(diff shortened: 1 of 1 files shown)


web/src/components/tables/RequestsFilterMenu.tsx
```diff
@@ -37,6 +37,18 @@ export const RequestsFilterMenu: React.FC<{
           defaultValue="all"
           title="View option"
           type="radio"
+          value={
+            // map the status input to a value so that the MenuList hydrates the current state
+            status === "PENDING"
+              ? "pend"
+              : status === "DECLINED"
+              ? "den"
```
chrnorm, line 45: maybe a typo here and should be `dec`? I see it's elsewhere in this file though

chrnorm (approved):

PR #477, opened 2023-02-15
Add link force command

meyerjrr:
## Describe your changes

- Adds ability to force a group deployment link to a group
- Quick refactor to move non api call out of api file and into TargetGroupService for listing target groups

## Type of change

## Issue and Documentation

-

## Testing

1.


pkg/api/targetgroup.go
```diff
@@ -11,45 +10,16 @@ import (
 	"github.com/common-fate/ddb"
 )
 
-func (a *API) FetchTargetGroups(ctx context.Context) []types.TargetGroup {
```
JoshuaWilkes, line 14: @meyerjrr Our service layer should be responsible for business logic,
OUr API layer is responsible for data fetching, serialising and calling service layer methods

In this PR you have shifted data fetching code from the API to the service layer, this goes against our desired separation of concerns here and abstracts the data fetching from the API

The way that this code handles no items returned and the fact that it packages them into an API type is not a concern of the service layer.

For this PR you should revert removing this fetch resources function and instead keep data fetching as an operation of the API layer

JoshuaWilkes (approved):

PR #479, opened 2023-02-15
added build time urls for registry and cf api to remove hardcoded values

meyerjrr:
Removes hard coded local api values for build time urls

### DIFF
--- a/cf/cmd/cli/commands/deployment/deployment.go
+++ b/cf/cmd/cli/commands/deployment/deployment.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 
 	"github.com/common-fate/clio"
+	"github.com/common-fate/common-fate/internal/build"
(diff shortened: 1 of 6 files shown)


chrnorm:
@meyerjrr you can replace the CF API environment variable with the following (see this [implementation in Granted](https://github.com/common-fate/granted/blob/main/pkg/granted/sso.go#L211-L219))

```go
import cfconfig "github.com/common-fate/cli/pkg/config"
import "github.com/common-fate/cli/pkg/client"

cfg, err := cfconfig.Load()
if err != nil {
  return nil, err
}

cf, err := client.FromConfig(c.Context, cfg) // this is automatically configured with the Common Fate API URL
```

Here's a video with a runthrough: https://www.loom.com/share/f310f34b9ef7481189c3152fd93761f0

PR #488, opened 2023-02-17
CF-694: health check, cache sync (cron variable)

jordiup:
## Describe your changes

Keen for a diligent review on this one, haven't worked in the cloud formation for a while. It's a straightforward change but a few variables to track across the multiple files/inputs. 

image

After running a fresh `mage deploy:dev` it seems that the event bridge CRON has been removed indicating our DE for the ticket

## Type of change

## Issue and Documentation
- https://linear.app/common-fate/issue/CF-694/add-the-ability-to-enable-the-cron-healthcheck-and-cache-sync-via-a

## Testing

1.

### DIFF
--- a/deploy/infra/bin/common-fate.ts
+++ b/deploy/infra/bin/common-fate.ts
@@ -40,6 +40,16 @@ const analyticsDeploymentStage = app.node.tryGetContext(
   "analyticsDeploymentStage"
 );
 
+let shouldRunCronHealthCheckCacheSync = app.node.tryGetContext(
+  "enableCronHealthCheck"
+);
+
+if (typeof shouldRunCronHealthCheckCacheSync === "string") {
+  if (shouldRunCronHealthCheckCacheSync.toLowerCase() == "true") {
+    shouldRunCronHealthCheckCacheSync = true;
+  }
+}
+
 // https://github.com/aws/aws-cdk/issues/11625
 // cdk processes both stacks event if you specify only one
 // To prepare the prod stack only, set the env var to "prod"
@@ -84,6 +94,8 @@ if (stackTarget === "dev") {
     analyticsUrl: analyticsUrl || "",
     analyticsLogLevel: analyticsLogLevel || "",
     analyticsDeploymentStage: analyticsDeploymentStage || "",
+    shouldRunCronHealthCheckCacheSync:
+      shouldRunCronHealthCheckCacheSync || false,
   });
 } else if (stackTarget === "prod") {
   new CommonFateStackProd(app, "Granted", {
--- a/deploy/infra/lib/constructs/cache-sync.ts
+++ b/deploy/infra/lib/constructs/cache-sync.ts
@@ -38,18 +38,17 @@ export class CacheSync extends Construct {
 
     props.dynamoTable.grantReadWriteData(this._lambda);
 
-    if (props.shouldRunAsCron) {
-      //add event bridge trigger to lambda
-      this.eventRule = new events.Rule(this, "EventBridgeCronRule", {
-        schedule: events.Schedule.cron({ minute: "0/5" }),
-      });
+    //add event bridge trigger to lambda
+    this.eventRule = new events.Rule(this, "EventBridgeCronRule", {
+      schedule: events.Schedule.cron({ minute: "0/5" }),
+      enabled: props.shouldRunAsCron,
+    });
 
-      // add the Lambda function as a target for the Event Rule
-      this.eventRule.addTarget(new targets.LambdaFunction(this._lambda));
+    // add the Lambda function as a target for the Event Rule
+    this.eventRule.addTarget(new targets.LambdaFunction(this._lambda));
 
-      // allow the Event Rule to invoke the Lambda function
-      targets.addLambdaPermission(this.eventRule, this._lambda);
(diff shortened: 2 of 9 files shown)


magefiles/magefile.go
```diff
@@ -360,6 +360,7 @@ func Dotenv() error {
 	myEnv["COMMONFATE_ACCESS_REMOTE_CONFIG_URL"] = cfg.Deployment.Parameters.ExperimentalRemoteConfigURL
 	myEnv["COMMONFATE_REMOTE_CONFIG_HEADERS"] = cfg.Deployment.Parameters.ExperimentalRemoteConfigHeaders
 	myEnv["COMMONFATE_GRANTER_V2_STATE_MACHINE_ARN"] = o.GranterV2StateMachineArn
+	myEnv["COMMONFATE_ENABLE_CRON_HEALTH_CHECK_IN_DEV"] = cfg.Deployment.Parameters.EnableCronHealthCheckInDev
```
JoshuaWilkes: I don't believe that we need this environment variable as the configuration comes from the deployment.yml file

```suggestion
```

pkg/deploy/config.go
```diff
@@ -511,6 +512,12 @@ func (c *Config) CfnParams() ([]types.Parameter, error) {
 			ParameterValue: &p.AnalyticsDeploymentStage,
 		})
 	}
+	if c.Deployment.Parameters.EnableCronHealthCheckInDev != "" {
```
JoshuaWilkes: This step/function is designed to generate cloudformation Parameters which are used to configure production deployments.

Because this configuration enableCronHealthCheckInDev is intended for development, it is passed into cdk as a flag directly.

You have implemented the cdk flag above correctly which will be used when running `mage deploy:dev`

When we add configuration that is used in both development and and Production deployments, then we would add a cloudformation parameter to the infrastructure stack, and generate a value for it using this method here.

```suggestion
```

deploy/infra/lib/constructs/cache-sync.ts
```diff
@@ -37,16 +38,19 @@ export class CacheSync extends Construct {
 
     props.dynamoTable.grantReadWriteData(this._lambda);
 
-    //add event bridge trigger to lambda
-    this.eventRule = new events.Rule(this, "EventBridgeCronRule", {
-      schedule: events.Schedule.cron({ minute: "0/5" }),
-    });
+    if (props.shouldRunAsCron) {
```
JoshuaWilkes: For both of these rules in cache sync and healthcheck, the if statement means that the rule will not be created in AWS at all when it is disabled.

The achieves the desired effect of being able to configure it, however the downside is that if you were doing development on the lambda cache sync, you need to do a stack deployment to turn syncing on.

This can take 5 - 6 minutes or so.

The alternative is that you can create the rule in CDK and just set it to disabled.
The benefit of this is that as a developer, I can go into the AWS console and just toggle the rule on or off if and when I need to, and I don't need to wait 5 minutes.

Using the config in deployment.yml I can enable the rule permanently.

```
      //add event bridge trigger to lambda
      schedule: events.Schedule.cron({ minute: "0/5" }),	      this.eventRule = new events.Rule(this, "EventBridgeCronRule", {
    });	        schedule: events.Schedule.cron({ minute: "0/5" }),
enabled: props.shouldRunAsCron,
      });

      // add the Lambda function as a target for the Event Rule
      this.eventRule.addTarget(new targets.LambdaFunction(this._lambda));


    // add the Lambda function as a target for the Event Rule	      // allow the Event Rule to invoke the Lambda function
    this.eventRule.addTarget(new targets.LambdaFunction(this._lambda));	      targets.addLambdaPermission(this.eventRule, this._lambda);
```

deploy/infra/bin/common-fate.ts
```diff
@@ -40,6 +40,16 @@ const analyticsDeploymentStage = app.node.tryGetContext(
   "analyticsDeploymentStage"
 );
 
+let shouldRunCronHealthCheckCacheSync = app.node.tryGetContext(
+  "enableCronHealthCheck"
+);
+
+if (typeof shouldRunCronHealthCheckCacheSync === "string") {
```
JoshuaWilkes, line 47: I believe in typescript/js you can use something like
```
someString = "true"
enabled = !!someString
```

which will cast the string to boolean

JoshuaWilkes (changes requested):
Good work on this @jordiup 

I have added some useful context, overall this looks good

deploy/infra/lib/constructs/cache-sync.ts
jordiup: Cheers Josh, solid feedback

pkg/deploy/config.go
jordiup: Thanks for the explainer here 👌

deploy/infra/bin/common-fate.ts
jordiup: image
🤧

jordiup: Not quite the neatest formatting but I think it covers the check, it also will covers `True` -> `true`. Happy to alter if you have any other suggestions?

JoshuaWilkes (approved):

PR #497, opened 2023-02-21
SOL-44: remote config types update

jordiup:
## Describe your changes

This is a downstream change for https://github.com/common-fate/cloud/pull/2 

Most importantly it updates `pkg/remoteconfig/openapi.yml` to include the `samlSSOMetadataURL` field


chrnorm:
@jordiup we'll manually do this update, so don't worry about reading this from remote config!

PR #506, opened 2023-02-24
Jordi/sol 34 importing only selected idp groupsusers

jordiup:
## Describe your changes

importing only selected IDP Groups/Users based on a predefined filter 

```diff
version: 2
deployment:
  parameters:
    AdministratorGroupID: '123567'
    IdentityProviderType: one-login
+    IdentityGroupFilter: admins|granted-admins
```

## Type of change

## Issue and Documentation

https://linear.app/common-fate/issue/SOL-34/importing-only-selected-idp-groupsusers

## Testing

1. Unit tests added
2. See linear ticket for demo loom

- [x] I have added tests that prove my fix is effective or that my feature works.


pkg/api/users.go
```diff
@@ -32,6 +32,7 @@ func (a *API) AdminListUsers(w http.ResponseWriter, r *http.Request, params type
 		return
 	}
 
+	// for some reason, jack@commonfate.io isn't bein return here :(((
```
chrnorm:
```suggestion
```

pkg/deploy/output.go
```diff
@@ -50,6 +50,7 @@ type Output struct {
 	IDPSyncExecutionRoleARN       string `json:"IDPSyncExecutionRoleARN"`
 	CacheSyncFunctionName         string `json:"CacheSyncFunctionName"`
 	CLIAppClientID                string `json:"CLIAppClientID"`
+	IdentityGroupFilter           string `json:"IdentityGroupFilter"`
```
chrnorm: We can be clearer that we're filtering names, rather than Group IDs here

```suggestion
	IdentityGroupNameFilter           string `json:"IdentityGroupFilter"`
```

chrnorm:
Looks good @jordiup. Initially I thought that you had introduced a `|` separator but I forgot that that was built in to regex matching!

My initial thinking was that we'd use an array here:
```diff
version: 2
deployment:
  parameters:
    AdministratorGroupID: '123567'
    IdentityProviderType: one-login
+    IdentityGroupFilter:
+      - admins
+      - devops 
```
but given that you can construct a regex to match multiple things I think your implementation is fine. One minor naming suggestion and then good to be merged

jordiup:
> but given that you can construct a regex to match multiple things I think your implementation is fine. One minor naming suggestion and then good to be merged

Cheers @chrnorm, think we should be good to go 👌

JoshuaWilkes (changes requested):
The env var `COMMONFATE_IDENTITY_GROUP_FILTER` needs to be added to the lambda definitions in the infrastructure for the rest api and the identity syncer 

Otherwise this will only work locally

You should make this update then deploy and test that the identity sync is filtering as expected in your deployed identity syncer

PR #514, opened 2023-02-27
Delete handler

meyerjrr:
## Describe your changes

- Adds delete endpoint for handlers. Fix return type on get endpoint to be for a single type rather than list 

## Type of change

## Issue and Documentation

-

## Testing

1.


pkg/api/handler.go
```diff
@@ -70,3 +70,25 @@ func (a *API) AdminGetHandler(w http.ResponseWriter, r *http.Request, id string)
...
+		return
+	}
+
+	err = a.DB.Delete(ctx, q.Result)
+	if err != nil {
+		apio.Error(ctx, w, err)
+		return
+	}
+
+	apio.JSON(ctx, w, q.Result.ToAPI(), http.StatusOK)
```
JoshuaWilkes:
```suggestion
	apio.JSON(ctx, w, nil, http.StatusOK)
```

openapi.yml
```diff
@@ -1154,13 +1154,30 @@ paths:
...
+      responses:
+        "200":
+          description: OK
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/TGHandler"
+        "401":
+          $ref: "#/components/responses/ErrorResponse"
+        "403":
```
JoshuaWilkes:
```suggestion
        "404":
```

openapi.yml
```diff
@@ -1154,13 +1154,30 @@ paths:
...
         name: id
         in: path
         required: true
+    delete:
+      summary: ""
+      operationId: admin-delete-handler
+      responses:
+        "200":
+          description: OK
+          content:
```
JoshuaWilkes, line 1170: Don't return a body for this API

JoshuaWilkes (changes requested):
I have requested that the API doesn't return a body on deletion and you also need to add API layer tests for basic functionality

openapi.yml
meyerjrr: @JoshuaWilkes Done, and looked into the correct response to use for delete (with no content) and most say to use 204 response. So have updated to use that.

JoshuaWilkes (approved):

PR #557, opened 2023-03-08
Fe health check

meyerjrr:
## Describe your changes

- Adds healthcheck button in frontend to run a healthcheck for all lambdas
Screenshot 2023-03-08 at 4 42 06 pm

## Type of change

## Issue and Documentation

-

## Testing

1.


pkg/api/healthcheck.go
```diff
@@ -0,0 +1,71 @@
...
+	"github.com/aws/aws-sdk-go-v2/service/lambda"
+	"github.com/common-fate/apikit/apio"
+	"github.com/common-fate/apikit/logger"
+	"github.com/common-fate/common-fate/pkg/cfaws"
+	"github.com/common-fate/common-fate/pkg/deploy"
+	"github.com/common-fate/common-fate/pkg/storage"
+)
+
+func (a *API) AdminRunHealthcheck(w http.ResponseWriter, r *http.Request) {
+	ctx := context.Background()
```
JoshuaWilkes: The healtcheck is implemented as a service, so instead of invoking the healthcheck lambda function, it would be better here to add the hgeathcheck service to the API struct

then

```
err := a.Healthcheck.Check(ctx)
```

You should ensure that the rest API has IAM permission to invoke the Handlers. 
Be sure you test this in your live dev deployment as well.

pkg/api/healthcheck.go
```diff
@@ -0,0 +1,71 @@
...
+
+	log := logger.Get(ctx)
+	dc, err := deploy.LoadConfig(deploy.DefaultFilename)
```
JoshuaWilkes: General feedback, You cannot use deploy.LoadConfig in this context because the deployment.yml file is not available in the lambda runtime.

pkg/api/healthcheck.go
```diff
@@ -0,0 +1,71 @@
...
+	if err != nil {
+		apio.Error(ctx, w, err)
+		return
+	}
+	o, err := dc.LoadOutput(ctx)
```
JoshuaWilkes: Instead of using load output, in the context of the API you should have a parameter which is initialised at startup, it would be loaded from the environment of teh lambda.

In the infrastructure, you add this environment variable and its value.

We use pkg/config to define the environment variables for the servers/lambda

pkg/api/healthcheck.go
```diff
@@ -0,0 +1,71 @@
...
+		log.Info("Successfully synced the healthcheck")
+	} else {
+		apio.Error(ctx, w, fmt.Errorf("healthcheck sync failed with lambda invoke status code: %d", res.StatusCode))
+		return
+	}
+
+	// now run a fetch
+	listHandlers := storage.ListHandlers{}
+
+	_, err = a.DB.Query(ctx, &listHandlers)
```
JoshuaWilkes: What was this query for?

JoshuaWilkes (changes requested):

JoshuaWilkes (approved):

PR #562, opened 2023-03-09
Add validation to create targetgroup & create handler request body

Eddie023:
## Describe your changes
1. Add validation to create targetgroup & create handler request body. 
2.  Id in both case is a valid alphanumeric value based on the regex `[-a-zA-Z0-9]*$\`

## Type of change


JoshuaWilkes (approved):

JoshuaWilkes (changes requested):
Could you please add tests for this PR

PR #584, opened 2023-03-27
Return created Access Requests in response

chrnorm:


pkg/api/request.go
```diff
@@ -244,11 +245,15 @@ func (a *API) UserCreateRequest(w http.ResponseWriter, r *http.Request) {
...
 	var res types.CreateRequestResponse
 
+	var requestIDs []string
 	for _, r := range result {
 		res.Requests = append(res.Requests, r.Request.ToAPI())
+		requestIDs = append(requestIDs, r.Request.ID)
 	}
 
-	apio.JSON(ctx, w, nil, http.StatusOK)
+	clio.Infow("created access requests", "request_ids", requestIDs)
```
JoshuaWilkes, line 254: api logger should be used in this context

```suggestion
        log := logger.Get(ctx)
	log.Infow("created access requests", "request_ids", requestIDs)
```

PR #593, opened 2023-05-11
Fix timing

meyerjrr:
### What changed?
- Created a create request separate from the grant type we save in the database, that uses iso8601 time for start and end of grants to preserve timezones when requesting access.
- Update event handler to allow invoke of provider lambdas
- Use granter in live workflow


pkg/service/workflowsvc/service.go
```diff
@@ -2,22 +2,91 @@ package workflowsvc
...
+	// request reviewers are users who have one or more groups to review on the request as a whole
+	RequestReviewers []string `json:"requestReviewers" dynamodbav:"requestReviewers, set"`
+}
+
+type WorkflowGrant struct {
+	// The user email
+	Subject string                               `json:"subject" dynamodbav:"subject"`
+	Status  types.RequestAccessGroupTargetStatus `json:"status" dynamodbav:"status"`
+	//the time which the grant starts
+	Start iso8601.Time `json:"start" dynamodbav:"start"`
```
JoshuaWilkes: @meyerjrr is the only difference here that the grant has iso times?
I think it was probably just an oversight from me not using the iso package on the access.Grant struct.
Would you be able to just update that struct instead and not have the duplicate here?

JoshuaWilkes (changes requested):

pkg/service/workflowsvc/service.go
meyerjrr: Tried that at first, but would save as nil to the dynamo table when saving the grant. Think this is why we followed this pattern in the old request flow?

JoshuaWilkes: oh, this is possibly because the type doesn't implement a ddb marshalling method

pkg/service/workflowsvc/service.go
```diff
@@ -112,8 +114,12 @@ func (s *Service) Revoke(ctx context.Context, requestID string, groupID string,
 			return ErrGrantInactive
 		}
 
+		zap.S().Infow("Can revoke. calling runtime revoke.")
```
JoshuaWilkes:
```suggestion
log := logger.Get(ctx)
		log.Infow("Can revoke. calling runtime revoke.")
```

pkg/access/group_target.go
```diff
@@ -44,10 +45,10 @@ type Grant struct {
 	Subject string                               `json:"subject" dynamodbav:"subject"`
 	Status  types.RequestAccessGroupTargetStatus `json:"status" dynamodbav:"status"`
 	//the time which the grant starts
-	Start time.Time `json:"start" dynamodbav:"start"`
+	Start iso8601.Time `json:"start" dynamodbav:"start"`
 	//the time the grant is scheduled to end
-	End          time.Time `json:"end" dynamodbav:"end"`
-	Instructions *string   `json:"instructions" dynamodbav:"instructions"`
+	End          iso8601.Time `json:"end" dynamodbav:"end"`
+	Instructions *string      `json:"instructions" dynamodbav:"instructions"`
```
JoshuaWilkes: Access instructions were intended to be saved as a seperate type in dynamo db, there is already an instructions data model, the idea was that the FE would fetch instructions seperately to the request itself

JoshuaWilkes (changes requested):

jordiup (approved):

JoshuaWilkes (approved):

PR #595, opened 2023-05-16
fix check if user has access to rule when grouping + test

meyerjrr:
### What changed?
 - Fixed bug where users were not checked if they had access to the rule when grouping targets.

### DIFF
--- a/pkg/service/preflightsvc/service.go
+++ b/pkg/service/preflightsvc/service.go
@@ -98,21 +98,19 @@ func (s *Service) ProcessPreflight(ctx context.Context, user identity.User, pref
 	return &preflight, nil
 }
 
-func (s *Service) UserCanAccessRule(ctx context.Context, userId string, a rule.AccessRule) (bool, error) {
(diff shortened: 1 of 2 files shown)


pkg/service/preflightsvc/service.go
```diff
@@ -98,7 +98,25 @@ func (s *Service) ProcessPreflight(ctx context.Context, user identity.User, pref
 	return &preflight, nil
 }
 
-func (s *Service) GroupTargets(ctx context.Context, targets []cache.Target) ([]access.PreflightAccessGroup, error) {
+func (s *Service) UserCanAccessRule(ctx context.Context, userId string, a rule.AccessRule) (bool, error) {
```
JoshuaWilkes: @meyerjrr I think you can simplify the check here, because access rules have a list of gropu ids which can access.
and the group targets method takes an identity.User as an input.

The user.groups list will have the group ids.
you can directly compare the users groups to the access rule groups to filter them

jordiup (approved):

PR #600, opened 2023-05-17
Access templates

meyerjrr:
### What changed?
Adds ability for users to create access request templates for the rest of their team


jordiup (approved):
Looks good

PR #601, opened 2023-05-17
Cf 1025 implement resource filter

jordiup:


pkg/target/group.go
```diff
@@ -19,7 +36,7 @@ type Group struct {
 	From From `json:"from" dynamodbav:"from"`
 
 	// Schema is denomalised and saved here for efficiency
-	Schema providerregistrysdk.Target `json:"schema" dynamodbav:"schema"`
+	Schema GroupSchema `json:"schema" dynamodbav:"schema"`
```
meyerjrr, line 39: How come we changed this type here? 
I think we should keep it as `providerregistrysdk.Target` and update the type in the providerregistrysdk repo

Eddie023: This was necessary because the `providerregistrysdk.Target` type doesn't contain `resourceSchema` which is necessary for Group Schema so that the single API call will contain the resourceSchema properties associated with the Target Argument.

meyerjrr (approved):

PR #604, opened 2023-05-18
Default duration

meyerjrr:
### What changed?
Adds default duration to access rule+ use default duration for requests
- Small fix to formatting time on slack notifications
- Small fix to audit log when errors occur


jordiup (approved):
Looks good. See slack message re: pnpm --dev

PR #616, opened 2023-08-02
Fix pagination for large datasets

meyerjrr:
### What changed?
Fixes a bug where pagination repeats pages for larger datasets

### DIFF
--- a/web/src/utils/usePaginatorApi.ts
+++ b/web/src/utils/usePaginatorApi.ts
@@ -90,12 +90,12 @@ export const usePaginatorApi = <
       setNextToken(data.next);
       // Otherwise incrementing internally between pages
     } else {
-      const tokenIfPresent = tokenStack[pageIndex + 1];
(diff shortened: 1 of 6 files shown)


web/src/utils/usePaginatorApi.ts
```diff
@@ -90,12 +90,12 @@ export const usePaginatorApi = <
...
-        setNextToken(tokenStack[pageIndex + 1]);
-      } else {
-        setNextToken(data.next);
-      }
+      // const tokenIfPresent = tokenStack[pageIndex + 1];
+      // if (tokenIfPresent) {
+      //   setNextToken(tokenStack[pageIndex + 1]);
+      // } else {
+      setNextToken(data.next);
+      // }
```
meyerjrr, line 98: This was the offending code. It was pulling old pages from the tokenStack. 

Commenting this out and using the page from the api fixes the issue.

Issue #620, opened 2023-09-22
Okta integration: Sync of users does too many API calls and too slow in accounts with lots of users


keymon:
# What?

Our Okta account has ~5k users. We noticed that the sync process is not able to complete, not even within the maximum timeout of 15 minutes.

Checking the okta implementation [here](https://github.com/common-fate/common-fate/blob/main/pkg/identity/identitysync/okta.go) we noticed that:
 - it lists ALL the users in Okta. There is no way to restrict this or filter it out.
 - It processes each user, calling the API to reads all the group it is member of
 - finally lists all the groups

This causes a lot of calls to the Okta API, consuming all the quotas and limits, and the process is unable to finish on time. 

The implementation later relies `IdentityGroupFilter` to filter out the groups, but that is after reading everything, so it is not possible to leverage it.

# Expected behaviour

We would expect:
 - some way to filter the users to sync. For instance, users belonging to some given group or with some attribute. 
 - Some way to filter the groups to sync, possibly using `IdentityGroupFilter` before calling the API sync
 - less API calls to find out the group membership of each user. Possibly querying the members per group, rather than groups per member.
  - some speedup implementing some parallel execution
  
# Proposals

@chrnorm comments:
> I think the best long term approach to solving this would be to have an endpoint which is [SCIM](https://developer.okta.com/docs/concepts/scim/) compatible, that way the sync is controlled on the IDP side and it should mitigate the timeout

keymon:
Closing this as reslved in #625 

Pending implement SCIM or so

PR #625, opened 2023-09-25
Add groups option to IdentityConfiguration to fetch only desired users

IvanVan:
### What changed?
This PR adds an optional field `groups` for `IdentityConfiguration` config. 

### Why?
We faced Okta rate limits, because commonfate called `ListUserGroups` for every user in org. See for more details - https://commonfatecommunity.slack.com/archives/C03SG8PPPH7/p1695293237007429 and https://github.com/common-fate/common-fate/issues/620

### How did you test it?
We tested it in our local environment. Unit tests - TBD.

### Potential risks
It shouldn't affect configurations that have been created before, because we keep the old behaviour for configuration without `groups` field.

No

### DIFF
--- a/pkg/identity/identitysync/okta.go
+++ b/pkg/identity/identitysync/okta.go
@@ -20,14 +20,14 @@ type OktaSync struct {
 	client   *okta.Client
 	orgURL   gconfig.StringValue
 	apiToken gconfig.SecretStringValue
-	groups   gconfig.StringValue
+	groups   gconfig.OptionalStringValue
 }
 
 func (s *OktaSync) Config() gconfig.Config {
 	return gconfig.Config{
 		gconfig.StringField("orgUrl", &s.orgURL, "the Okta organization URL"),
 		gconfig.SecretStringField("apiToken", &s.apiToken, "the Okta API token", gconfig.WithNoArgs("/granted/secrets/identity/okta/token")),
-		gconfig.StringField("groups", &s.groups, "Groups that users are synced from."),
+		gconfig.OptionalStringField("groups", &s.groups, "Groups that users are synced from."),
 	}
 }
 
@@ -57,7 +57,7 @@ func (s *OktaSync) TestConfig(ctx context.Context) error {
 }
 
 // userFromOktaUser converts a Okta user to the identityprovider interface user type
-func (o *OktaSync) idpUserFromOktaUser(ctx context.Context, oktaUser *okta.User) (identity.IDPUser, error) {
+func (o *OktaSync) idpUserFromOktaUser(ctx context.Context, oktaUser *okta.User, userIdToGroupIds map[string][]string) (identity.IDPUser, error) {
 	log := logger.Get(ctx).With("okta.user_id", oktaUser.Id)
 
 	userJSON, err := json.Marshal(oktaUser)
@@ -97,12 +97,16 @@ func (o *OktaSync) idpUserFromOktaUser(ctx context.Context, oktaUser *okta.User)
 
 	log.Debug("listing groups for user")
 
-	userGroups, _, err := o.client.User.ListUserGroups(ctx, oktaUser.Id)
-	if err != nil {
-		return u, errors.Wrapf(err, "listing groups for okta user %s", oktaUser.Id)
-	}
-	for _, g := range userGroups {
-		u.Groups = append(u.Groups, g.Id)
+	if userIdToGroupIds != nil {
+		u.Groups = append(u.Groups, userIdToGroupIds[oktaUser.Id]...)
+	} else {
+		userGroups, _, err := o.client.User.ListUserGroups(ctx, oktaUser.Id)
+		if err != nil {
+			return u, errors.Wrapf(err, "listing groups for okta user %s", oktaUser.Id)
+		}
+		for _, g := range userGroups {
+			u.Groups = append(u.Groups, g.Id)
+		}
 	}
 
 	log.Debugw("finished converting okta user to internal user", "user", u)
@@ -128,7 +132,7 @@ func logResponseErr(log *zap.SugaredLogger, res *okta.Response, err error) {
 	}
 }
 
-func (o *OktaSync) ListUsers(ctx context.Context) ([]identity.IDPUser, error) {
+func (o *OktaSync) listAllUsers(ctx context.Context) ([]identity.IDPUser, error) {
 	log := logger.Get(ctx)
 
 	//get all users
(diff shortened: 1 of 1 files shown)


pkg/identity/identitysync/okta.go
```diff
@@ -19,12 +20,14 @@ type OktaSync struct {
...
 	orgURL   gconfig.StringValue
 	apiToken gconfig.SecretStringValue
+	groups   gconfig.StringValue
 }
 
 func (s *OktaSync) Config() gconfig.Config {
 	return gconfig.Config{
 		gconfig.StringField("orgUrl", &s.orgURL, "the Okta organization URL"),
 		gconfig.SecretStringField("apiToken", &s.apiToken, "the Okta API token", gconfig.WithNoArgs("/granted/secrets/identity/okta/token")),
+		gconfig.StringField("groups", &s.groups, "Groups that users are synced from."),
```
keymon: IMO, we shouldn't break the previous interface/contract, and make this optional.

Now I am not sure if `gconfig.StringField` makes this mandatory, but there is a suspiciously named `gconfig.OptionalStringField()`!

pkg/identity/identitysync/okta.go
```diff
@@ -133,28 +136,36 @@ func (o *OktaSync) ListUsers(ctx context.Context) ([]identity.IDPUser, error) {
...
-		return nil, errors.Wrap(err, "listing okta users from okta API")
-	}
-
-	log.Debugw("listed all okta users")
-
-	for res.HasNextPage() {
-		var nextUsers []*okta.User
-		res, err = res.Next(ctx, &nextUsers)
+	oktaUsers := make(map[string]*okta.User)
+	for _, groupId := range strings.Split(o.groups.Get(), ",") {
```
keymon, line 181: As commented, we should make this optional, so keep the old and the newer implementation. 

Funny, the page by page iterator has the same signature, in theory you could reuse that bit of code, but would suggest to duplicate for clarity

keymon: BTW, makes sense to use a list as the gconfigField object seems to only support strings. Lets keep it simple. Also, we will likely only have one group here.

pkg/identity/identitysync/okta.go
keymon: My second comment, shall we call this plainly `groups` or something more explicit like `user_groups`?

it is possible later we add a `groups` to refer to the groups we do want to sync.

pkg/identity/identitysync/okta.go
```diff
@@ -133,28 +136,36 @@ func (o *OktaSync) ListUsers(ctx context.Context) ([]identity.IDPUser, error) {
...
+		log.Infow(fmt.Sprintf("fetching users for groupId = %s", groupId))
+		users, res, err := o.client.Group.ListGroupUsers(ctx, groupId, &query.Params{})
 		if err != nil {
+			// try and log the response body
 			logResponseErr(log, res, err)
-			return nil, err
+			return nil, errors.Wrap(err, "listing okta users from okta API")
```
keymon, line 187: Not sure here if we shouldn't bailout immediately and error on the first error.

Alternatively, we could acummulate the errors and return the aggregated list. 

but for the sake of KISS, it is better this way.

pkg/identity/identitysync/okta.go
```diff
@@ -133,28 +136,36 @@ func (o *OktaSync) ListUsers(ctx context.Context) ([]identity.IDPUser, error) {
 
 	log.Debugw("listing all okta users")
```
keymon: This log is in the original code but weird to have it here. Can we move it after the pagination?

pkg/identity/identitysync/okta.go
IvanVan: Fixed, now it's optional.

IvanVan: > it is possible later we add a groups to refer to the groups we do want to sync.

what do you mean? currently these are the groups that we want to sync users from, right?

pkg/identity/identitysync/okta.go
IvanVan: Fixed.

pkg/identity/identitysync/okta.go
IvanVan: Why? Looks like it indicates that listing process has been started which makes sense to keep it here.

pkg/identity/identitysync/okta.go
chrnorm: Thanks for the note here @keymon. Agreed on failing early for KISS and also to retain consistency with our other IDP integrations which have this behaviour. In practice the most common failure mode we've observed is total failure of the IDP sync component due to misconfigured API keys

pkg/identity/identitysync/okta.go
```diff
@@ -165,6 +172,59 @@ func (o *OktaSync) ListUsers(ctx context.Context) ([]identity.IDPUser, error) {
 	return idpUsers, nil
 }
 
+func (o *OktaSync) listUsersBasedOnGroups(ctx context.Context) ([]identity.IDPUser, error) {
```
keymon, line 175: nit?
```suggestion
func (o *OktaSync) listUsersByGroups(ctx context.Context) ([]identity.IDPUser, error) {
```

pkg/identity/identitysync/okta.go
```diff
@@ -19,12 +20,14 @@ type OktaSync struct {
...
 	orgURL   gconfig.StringValue
 	apiToken gconfig.SecretStringValue
+	groups   gconfig.OptionalStringValue
 }
 
 func (s *OktaSync) Config() gconfig.Config {
 	return gconfig.Config{
 		gconfig.StringField("orgUrl", &s.orgURL, "the Okta organization URL"),
 		gconfig.SecretStringField("apiToken", &s.apiToken, "the Okta API token", gconfig.WithNoArgs("/granted/secrets/identity/okta/token")),
+		gconfig.OptionalStringField("groups", &s.groups, "Groups that users are synced from."),
```
keymon, line 30: I would rename this option, instead of just `groups`, something more explicit like `user_groups`.

In this implementation we are still reading all the groups in `ListGroups`, but getting the users 
 
```suggestion
		gconfig.OptionalStringField("user_groups", &s.userGroups, "Groups that users are synced from."),
```

pkg/identity/identitysync/okta.go
```diff
@@ -94,12 +97,16 @@ func (o *OktaSync) idpUserFromOktaUser(ctx context.Context, oktaUser *okta.User)
 
 	log.Debug("listing groups for user")
 
-	userGroups, _, err := o.client.User.ListUserGroups(ctx, oktaUser.Id)
-	if err != nil {
-		return u, errors.Wrapf(err, "listing groups for okta user %s", oktaUser.Id)
-	}
-	for _, g := range userGroups {
-		u.Groups = append(u.Groups, g.Id)
+	if userIdToGroupIds != nil {
```
keymon, line 100: nit: for readbility,  instead of having this function behaving in 2 different ways, I would do the logic for the groups in each loop of each `listAllUsers()` and `listUsersByGroups()`

pkg/identity/identitysync/okta.go
```diff
@@ -165,6 +172,59 @@ func (o *OktaSync) ListUsers(ctx context.Context) ([]identity.IDPUser, error) {
...
+			userGroupIds[user.Id] = append(userGroupIds[user.Id], groupId)
+		}
+	}
+
+	// convert all Okta users to internal users
+	for _, u := range oktaUsers {
+		user, err := o.idpUserFromOktaUser(ctx, u, userGroupIds)
+		if err != nil {
+			return nil, errors.Wrapf(err, "converting okta user %s to internal user", u.Id)
+		}
```
keymon, line 214: nit, As my other comment, put the logic of the group in the caller.

```suggestion
		user, err := o.idpUserFromOktaUser(ctx, u)
		if err != nil {
			return nil, errors.Wrapf(err, "converting okta user %s to internal user", u.Id)
		}
		user.Groups = userIdToGroupIds[u.Id]
```

chrnorm (approved):
Thankyou for the contribution @IvanVan and thankyou @keymon for the code review! This looks good to me so we'll get this upstreamed as a patch release.

pkg/identity/identitysync/okta.go
```diff
@@ -155,7 +162,7 @@ func (o *OktaSync) ListUsers(ctx context.Context) ([]identity.IDPUser, error) {
 
 	// convert all Okta users to internal users
 	for _, u := range users {
-		user, err := o.idpUserFromOktaUser(ctx, u)
+		user, err := o.idpUserFromOktaUser(ctx, u, nil)
```
keymon, line 165: nit, as per my other comment, move the different logic to the caller so `idpUserFromOktaUser` does only one thing and the behaviour does not change based on parameters.
```suggestion
		user, err := o.idpUserFromOktaUser(ctx, u)
		if err != nil {
			return nil, errors.Wrapf(err, "converting okta user %s to internal user", u.Id)
		}
		userGroups, _, err := o.client.User.ListUserGroups(ctx, u.Id)
		if err != nil {
			return u, errors.Wrapf(err, "listing groups for okta user %s", u.Id)
		}
		for _, g := range userGroups {
			user.Groups = append(user.Groups, g.Id)
		}
	}
```

PR #626, opened 2023-09-27
Add wildcard option to group field

IvanVan:
### What changed?
Follow up PR to https://github.com/common-fate/common-fate/pull/625
This PR adds functionality to support wildcard in `groups` field for IdentityConfiguration config. Thus common-fate fetches all the groups that Okta API token has access to.

#### Current state:

We do ListUsers = ListAllUsersAPICall ([code](https://github.com/common-fate/common-fate/blob/5b5b01647ed62d68fe5f6dffeccdd09306b2197e/pkg/identity/identitysync/okta.go#L136)) + numberOfUsers * GetUserGroupsAPICall ([code](https://github.com/common-fate/common-fate/blob/5b5b01647ed62d68fe5f6dffeccdd09306b2197e/pkg/identity/identitysync/okta.go#L97)) and we do ListGroups(ListAllGroupsAPICall). Every API call also may include pagination, but we don’t take it into account in these calculations. Also number of operations to parse/create map/etc. is O(numberOfGroups * numberOfUser.

#### This PR:
We'll call ListGroups first if the `groups` field is provided, then we will fetch all users for every group and then group membership will be used to fill `Groups` field for every user. In this case we need to fetch users for every groups. So in total it will be ListGroups(ListAllGroupsAPICall) + numberOfGroups * GetGroupUsers  + ListAllUsersAPICall (it may be omit, because we already have all the users from groups that commonfate has access to). Number of operations is the same O(numberOfGroups * numberOfUser).

Thus this optimization makes sense if `# groups << # users`. We'll add this information to docs here - https://github.com/common-fate/docs if the PR is approved.

### How did you test it?
We tested it in our env.

### Potential risks
N/A

No

### Link to relevant docs PRs
https://github.com/common-fate/common-fate/pull/625

### DIFF
--- a/pkg/identity/identitysync/okta.go
+++ b/pkg/identity/identitysync/okta.go
@@ -172,13 +172,56 @@ func (o *OktaSync) listAllUsers(ctx context.Context) ([]identity.IDPUser, error)
 	return idpUsers, nil
 }
 
+func (o *OktaSync) listGroups(ctx context.Context) ([]*okta.Group, error) {
+	log := logger.Get(ctx)
+
+	log.Debugw("listing all okta groups")
+
+	groups, res, err := o.client.Group.ListGroups(ctx, &query.Params{})
+	if err != nil {
+		// try and log the response body
+		logResponseErr(log, res, err)
+		return nil, errors.Wrap(err, "listing okta groups from okta API")
+	}
+
+	log.Debugw("listed all okta groups")
+
+	for res.HasNextPage() {
+		log.Debugw("hasmore")
+		var nextGroups []*okta.Group
+		res, err = res.Next(ctx, &nextGroups)
+		if err != nil {
+			logResponseErr(log, res, err)
+			return nil, err
+		}
+		groups = append(groups, nextGroups...)
+		log.Debugw("fetched more groups", "nextPage", res.NextPage)
+	}
+
+	return groups, nil
+}
+
 func (o *OktaSync) listUsersBasedOnGroups(ctx context.Context) ([]identity.IDPUser, error) {
 	log := logger.Get(ctx)
 	idpUsers := []identity.IDPUser{}
 
 	oktaUsers := make(map[string]*okta.User)
 	userGroupIds := make(map[string][]string)
-	for _, groupId := range strings.Split(o.groups.Get(), ",") {
+
+	var groupIds []string
+	if o.groups.Get() == "*" {
+		oktaGroups, err := o.listGroups(ctx)
+		if err != nil {
+			return nil, err
+		}
+		for _, oktaGroup := range oktaGroups {
+			groupIds = append(groupIds, oktaGroup.Id)
+		}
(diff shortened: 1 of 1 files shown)


shwethaumashanker:
@IvanVan Where you able to run `gdeploy identity sync` successfully after this update ? I'm seeing this error: "listing okta users from okta API: the API returned an error: Not found: Resource not found: 00g* (UserGroup): withStack
null" in the APIIdpSyncHandlerFunction

I tested using this format: `groups: 00g*`

keymon:
@shwethaumashanker actually the only valid options are static values or just `*`, `00g*` is not valid

the idea is that by having `groups: *` we are inversing the logic of the membership api queries, where the service will list all the groups accessible by the sync cluster, and check members for each.

This is great for our use case, because then we leverage the concept of ResourceSet in okta https://help.okta.com/en-us/content/topics/security/custom-admin-role/create-resource-set.htm to restrict which groups commonfate can see, and this way control a lot the API calls.

But yes, `groups: *` might be a bit confusing/misleading UX: 
 -  `groups` refers to "groups to get user membership from". IMO a better name could be `user_groups`
 - `groups:*` t effectivelly says "Sync membership from all accessible groups"
 - `groups: qwertyuio234,qwertyuio235,qwertyuio236` it effectivelly says "Sync membership from all accessible groups"
  

We are open to alternatives. Some options:
 - Option 1: `groups: *` `groups: qwertyuio234,qwertyuio235,qwertyuio236`  
 - Option 2: `groups: any` `groups: qwertyuio234,qwertyuio235,qwertyuio236`  
 - Option 3: `membership_from_groups: true` + `groups: qwertyuio234,qwertyuio235,qwertyuio236`
    - if `users_from_groups: true` then pull members from groups
    - if `groups: qwertyuio234,qwertyuio235,qwertyuio236` set, only sync/query those groups

IvanVan:
> @IvanVan Where you able to run gdeploy identity sync successfully after this update ? I'm seeing this error: "listing okta users from okta API: the API returned an error: Not found: Resource not found: 00g* (UserGroup): withStack
> null" in the APIIdpSyncHandlerFunction
> 
> I tested using this format: groups: 00g*

Thank you @shwethaumashanker for reviewing!
As @keymon said there are only two available options in the current implementation: "*"(all the groups) and "groupId1,groupId2"(group ids separated by comma). We tested it in our environment with "*" option.

shwethaumashanker:
Hi @IvanVan and @keymon,

Thanks for the clarification – my bad for misunderstanding the options earlier. I tested `groups: "*"` and it seems to be working well. As long as we keep things well-documented, I think this implementation should do the trick. 

I'll released v0.15.9 with the PR, and for future PRs, we're sticking with the `release/v0.15` branch for OSS releases.

Appreciate your contributions – thank you both! 🙌

shwethaumashanker (approved):

PR #634, opened 2023-10-26
Add option to run lambda as autoapproval

IvanVan:
### What changed?
This PR adds an option to specify lambda that should be triggered when access request is created. The lambda returns if the request should be auto-approved.

### Why?
Because we need an option to auto-approve request for people who are oncall.

### How did you test it?
I tested it locally for dev env. I need to test in a cloud - TBD.

### Potential risks
No

### Link to relevant docs PRs
https://commonfatecommunity.slack.com/archives/C03SG8PPPH7/p1697666320666679


keymon:
Seeing the architecture of the glide, I wonder if a better pattern would be to implement apart a new lambda and subscribe to the events of eventbridge. similar to how the notifier and audit works:

image

- https://github.com/common-fate/glide/blob/release/v0.15/pkg/notifiers/slack/handle_request.go#L25
-  https://github.com/common-fate/glide/blob/release/v0.15/pkg/eventhandler/eventhandler.go#L27-L39
- I think that is activated by a event bridge rule, like defined here: https://github.com/common-fate/glide/blob/release/v0.15/deploy/infra/lib/constructs/notifiers.ts#L56-L65

I will let the commonfate team comment on this, what is a better approach.

chrnorm:
Great point here @keymon. My view is that a synchronous Lambda should work well for this use case as we don't yet have APIs that a service account could use to manipulate the state of a pending access request. Aside from this it keeps it nice and simple and shouldn't have too much of an impact on UX as long as your Lambda executes quickly.

deploy/infra/lib/common-fate-stack-dev.ts
```diff
@@ -67,9 +68,14 @@ export class CommonFateStackDev extends cdk.Stack {
       idpSyncTimeoutSeconds,
       idpSyncSchedule,
       idpSyncMemory,
+      autoApprovalLambdaARN,
     } = props;
     const appName = `common-fate-${stage}`;
 
+    console.log("!!!!")
+    console.log(autoApprovalLambdaARN)
+    console.log("!!!!")
```
chrnorm:
```suggestion
```

pkg/autoapproval/autoapproval.go
```diff
@@ -0,0 +1,79 @@
...
+}
+
+type Service struct {
+}
+
+func (s Service) Autoapprove(user identity.User, rule rule.AccessRule, lambdaArn string) (bool, error) {
+	sess, err := session.NewSession()
+	//	&aws.Config{
+	//	Region: aws.String("us-west-2"), // Replace with your region
+	//}
```
chrnorm:
```suggestion
```

chrnorm (approved):
Thankyou @IvanVan! 🙌

pkg/service/accesssvc/create.go
```diff
@@ -205,9 +207,18 @@ func (s *Service) createRequest(ctx context.Context, in createRequestOpts) (Crea
 	if err != nil {
 		return CreateRequestResult{}, err
 	}
+	// TODO(IvanVan) Get it from config
+	arn := os.Getenv("COMMONFATE_AUTO_APPROVAL_LAMBDA_ARN")
+	var autoapproved bool
+	if arn != "" {
+		autoapproved, err = autoapproval.Service{}.Autoapprove(in.User, in.Rule, "123")
```
keymon, line 214: Fixed here: https://github.com/common-fate/glide/pull/642/files#diff-e6133f6c0ef497adc2b8f4726520a08b12f07be2a999f850f8f8f3b83df4c676R214

PR #635, opened 2023-10-27
Add vpc to target granter lambda

IvanVan:
### What changed?
This PR introduce two new settings that users can provide `LambdaSubnetIds` and `LambdaSecurityGroups`. Only if two of them are provided changes take place.
 `LambdaSubnetIds` - list of subnets ids that are used by all lambdas
`LambdaSecurityGroups` - list of security groups that are used by all lambdas

### Why?
This is necessary due to security requirements on the IP list from where our resources(such as Okta groups) can be modified/accessed

### How did you test it?
1. Development stack was deployed to devel environment by running "mage deploy:dev"
2. Production stack was synthesized by running `mage release:publishCloudFormation  bucket-name hash`

### Potential risks
No

Yes

### Link to relevant docs PRs
https://github.com/common-fate/glide/issues/623

### DIFF
--- a/deploy/infra/lib/constructs/targetgroup-granter.ts
+++ b/deploy/infra/lib/constructs/targetgroup-granter.ts
@@ -7,8 +7,7 @@ import * as path from "path";
 import { EventBus } from "aws-cdk-lib/aws-events";
 import { grantAssumeHandlerRole } from "../helpers/permissions";
 import { Table } from "aws-cdk-lib/aws-dynamodb";
-import * as ec2 from 'aws-cdk-lib/aws-ec2'
-import {BaseLambdaFunction} from "../helpers/base-lambda";
+import { BaseLambdaFunction } from "../helpers/base-lambda";
 
 interface Props {
   eventBusSourceName: string;
@@ -32,7 +31,6 @@ export class TargetGroupGranter extends Construct {
         "targetgroup-granter.zip"
       )
     );
-
 
     this._lambda = new BaseLambdaFunction(this, "StepHandlerFunction", {
       functionProps: {
--- a/deploy/infra/lib/constructs/targetgroup-granter.ts
+++ b/deploy/infra/lib/constructs/targetgroup-granter.ts
@@ -7,13 +7,13 @@ import * as path from "path";
 import { EventBus } from "aws-cdk-lib/aws-events";
 import { grantAssumeHandlerRole } from "../helpers/permissions";
 import { Table } from "aws-cdk-lib/aws-dynamodb";
-import { BaseLambdaFunction } from "../helpers/base-lambda";
+import { BaseLambdaFunction, VpcConfig } from "../helpers/base-lambda";
 
 interface Props {
   eventBusSourceName: string;
   dynamoTable: Table;
   eventBus: EventBus;
-  vpcConfig: any;
+  vpcConfig: VpcConfig;
 }
 export class TargetGroupGranter extends Construct {
   private _stateMachine: sfn.StateMachine;
--- a/deploy/infra/lib/constructs/targetgroup-granter.ts
+++ b/deploy/infra/lib/constructs/targetgroup-granter.ts
@@ -7,11 +7,13 @@ import * as path from "path";
 import { EventBus } from "aws-cdk-lib/aws-events";
 import { grantAssumeHandlerRole } from "../helpers/permissions";
 import { Table } from "aws-cdk-lib/aws-dynamodb";
+import * as ec2 from 'aws-cdk-lib/aws-ec2'
 
 interface Props {
   eventBusSourceName: string;
   dynamoTable: Table;
   eventBus: EventBus;
+  lambdaVpcId: string,
 }
 export class TargetGroupGranter extends Construct {
   private _stateMachine: sfn.StateMachine;
@@ -30,6 +32,10 @@ export class TargetGroupGranter extends Construct {
       )
     );
 
+    const vpc = ec2.Vpc.fromLookup(this, 'Vpc', {
+      vpcId: props.lambdaVpcId,
+    });
+
     this._lambda = new lambda.Function(this, "StepHandlerFunction", {
       code,
       timeout: Duration.minutes(5),
@@ -40,6 +46,7 @@ export class TargetGroupGranter extends Construct {
       },
       runtime: lambda.Runtime.GO_1_X,
       handler: "targetgroup-granter",
+      vpc: vpc,
     });
 
     props.dynamoTable.grantReadWriteData(this._lambda);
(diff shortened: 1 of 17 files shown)


deploy/infra/lib/constructs/targetgroup-granter.ts
```diff
@@ -7,11 +7,13 @@ import * as path from "path";
 import { EventBus } from "aws-cdk-lib/aws-events";
 import { grantAssumeHandlerRole } from "../helpers/permissions";
 import { Table } from "aws-cdk-lib/aws-dynamodb";
+import * as ec2 from 'aws-cdk-lib/aws-ec2'
```
keymon:
```suggestion
import { Vpc } as ec2 from 'aws-cdk-lib/aws-ec2'
```
To follow the pattern

deploy/infra/lib/constructs/targetgroup-granter.ts
```diff
@@ -30,6 +32,10 @@ export class TargetGroupGranter extends Construct {
       )
     );
 
+    const vpc = ec2.Vpc.fromLookup(this, 'Vpc', {
```
keymon:
```suggestion
    const vpc = Vpc.fromLookup(this, 'Vpc', {
```

keymon (changes requested):
The idea is good but caveat: we must specify vpc id, list of subnet ids, and ideally (possibly optionally) security group id.

I also think we must apply this to every lambda and resource

image

deploy/infra/lib/helpers/base-lambda.ts
```diff
@@ -0,0 +1,28 @@
...
+
+export interface BaseLambdaFunProps {
+  functionProps: lambda.FunctionProps;
+  vpcConfig: any;
+}
+
+export class BaseLambdaFunction extends lambda.Function {
+  constructor(scope: any, id: string, props: BaseLambdaFunProps) {
+    super(scope, id, props.functionProps);
+    this.addToRolePolicy(
```
keymon, line 21: What about attach the default policy from AWS AWSLambdaVPCAccessExecutionRole. https://docs.aws.amazon.com/aws-managed-policy/latest/reference/AWSLambdaVPCAccessExecutionRole.html 

it is almost the same

deploy/infra/lib/helpers/base-lambda.ts
```diff
@@ -0,0 +1,28 @@
+import * as cdk from "aws-cdk-lib";
+import * as lambda from "aws-cdk-lib/aws-lambda";
+import { PolicyStatement } from "aws-cdk-lib/aws-iam";
+
+export interface BaseLambdaFunProps {
+  functionProps: lambda.FunctionProps;
+  vpcConfig: any;
+}
```
keymon: nit and just curious, could this be an extend? 

```suggestion
export interface BaseLambdaFunProps extends lambda.FunctionProps{
  vpcConfig: any;
}
```

but does not matter at all, looks great

keymon (approved):
Cool! LGTM. ofc, up to upstream :)

deploy/infra/lib/helpers/base-lambda.ts
chrnorm, line 21: In cases where users aren't using VPCs with their deployment this will add unnecessary permissions - could you please attach [a `CfnCondition`](https://loige.co/create-resources-conditionally-with-cdk/#using-cfncondition-with-cdk) to this to only attach these permissions if the VPC is in use?

(unfortunately the way we render the CDK code into CloudFormation means that we can't do this at synthesis time which adds a bit of complexity. Usually you would just add some TypeScript code to conditionally add the PolicyStatement)

deploy/infra/lib/constructs/targetgroup-granter.ts
```diff
@@ -7,11 +7,13 @@ import * as path from "path";
 import { EventBus } from "aws-cdk-lib/aws-events";
 import { grantAssumeHandlerRole } from "../helpers/permissions";
 import { Table } from "aws-cdk-lib/aws-dynamodb";
+import { BaseLambdaFunction } from "../helpers/base-lambda";
 
 interface Props {
   eventBusSourceName: string;
   dynamoTable: Table;
   eventBus: EventBus;
+  vpcConfig: any;
```
chrnorm: nitpick: if you could strongly define this type it would be fantastic.

chrnorm (changes requested):
Looks good and just a minor requested change on conditionally adding the IAM policies to the lambda functions!

deploy/infra/lib/constructs/targetgroup-granter.ts
IvanVan: I've added two required fields for VpcConfig object - subnetIds and securityGroupIds. But these fields are still `any` type. Because the intended data type for the SubnetIds and SecurityGroupIds fields is a list of strings. However, in the production stack, the type of these fields is ICfnRuleConditionExpression, while in the development stack, it is string[].

deploy/infra/lib/helpers/base-lambda.ts
IvanVan: Great point @chrnorm! Thank you for pointing it out!

Regarding the link you provided: As far as I understand it is for L1 resources ([link](https://docs.aws.amazon.com/cdk/v2/guide/constructs.html). However in our case it's a property that we add to a resource(namely IAM role which is created by default if you don't provide it to a constructor of lambda object). As you said we can't use runtime if statements because we do synthesis.

Some facts: A lambda function has only one execution role. Role is a high level L1 resources in terms of CF, so we can override different properties of it using structure like this:
```
    const cfnDist = resource.node.defaultChild as cdk.CfnResource;
    cfnDist.addPropertyOverride("PropertyName", "PropertyVal");
```

Possible solution: we can refactor code and create an execution lambda role ourselves rather than calling `addToRolePolicy` method. In this case we can do some "if/else" constructions because IAM Policy is construct. 
```
// Create a new IAM role
const myRole = new iam.Role(this, 'MyRole', {
  assumedBy: new iam.ServicePrincipal('lambda.amazonaws.com')
});

// Create policy
const policyStatement = new iam.PolicyStatement({
  effect: iam.Effect.ALLOW,
  actions: [''],
  resources: ['']
});

// Do some if else manipulation

statements = cdk.Fn.conditionIf(attachLambdaToVpcCondition.logicalId,
{"actions if condition is true"},
{"actions if condition is false"})

const cfnDist = policyStatement.node.defaultChild as cdk.CfnResource;
 cfnDist.addPropertyOverride("PolicyDocument", statements);

// create a lambda with a role
const myFunction = new lambda.Function(this, 'MyFunction', {
  role: myRole
});
```
But it requires more or less a major refactor that we can do in the next PR. WDYT?

chrnorm (approved):
Thankyou @IvanVan!

PR #642, opened 2023-11-29
Fix autoapproval functionality and vpc config

IvanVan:
### What changed?
This is the follow up PR to https://github.com/common-fate/glide/pull/634 and https://github.com/common-fate/glide/pull/635

### Why?
1. It fixes issues that introduced here - https://github.com/common-fate/glide/pull/634/files#diff-e6133f6c0ef497adc2b8f4726520a08b12f07be2a999f850f8f8f3b83df4c676R214. It should be lambda arn instead of `123`.
2. If fixes a bug introduced in this commit - https://github.com/common-fate/glide/pull/635/commits/d5059c9a41d025a7724c92311d5a410f5226ae61. vpcConfig fields should start from capital letter otherwise error is thrown during deployment
```
extraneous key is not permitted
```

### How did you test it?
1. Development stack was deployed to devel environment by running "mage deploy:dev"
3. Production stack was synthesized by running mage release:publishCloudFormation  bucket-name hash

### Potential risks
Changes are scope to CDK logic so potential bugs should be isolated to deployment stage.

### DIFF
--- a/deploy/infra/bin/common-fate.ts
+++ b/deploy/infra/bin/common-fate.ts
@@ -108,7 +108,7 @@ if (stackTarget === "dev") {
     idpSyncMemory: idpSyncMemory || 128,
     idpSyncSchedule: idpSyncSchedule || "rate(5 minutes)",
     idpSyncTimeoutSeconds: idpSyncTimeoutSeconds || 30,
-    autoApprovalLambdaARN: autoApprovalLambdaARN,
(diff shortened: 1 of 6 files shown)


chrnorm (approved):
Thanks @IvanVan! Merging and I will run an additional test deploy to verify the fix and cut a new bugfix release for this

PR #647, opened 2023-12-05
revert attributevalue version

shwethaumashanker:
### What changed?
revert attributevalue version

### Why?
Fixes a 401 error on the web app launching


shwethaumashanker:
Still getting the same 401 error

